PR #1, opened 2020-03-25
Add ilcsoft steering file converter

petricm:


andresailer:
How about making this python3 compatible?

andresailer:
Thanks!

PR #19, opened 2020-10-06
Added test for converter script and CLIC reconstruction

fdplacido:
BEGINRELEASENOTES
- Added test which invokes the XML to Python converter
- The same test runs and checks the complete clicReconstruction

ENDRELEASENOTES

### DIFF
--- a/.github/scripts/gmp_build_and_test.sh
+++ b/.github/scripts/gmp_build_and_test.sh
@@ -15,4 +15,4 @@ cmake \
   .. && \
 make -j 2 && \
 make install && \
-ctest --output-on-failure
+ctest --verbose --output-on-failure
--- a/.github/scripts/gmp_build_and_test.sh
+++ b/.github/scripts/gmp_build_and_test.sh
@@ -15,4 +15,6 @@ cmake \
   .. && \
 make -j 2 && \
 make install && \
+ctest -R test_gmp1 -V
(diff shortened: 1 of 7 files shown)


.github/scripts/gmp_build_and_test.sh
```diff
@@ -15,4 +15,6 @@ cmake \
   .. && \
 make -j 2 && \
 make install && \
+ctest -R test_gmp1 -V
```
andresailer, line 18: why not just 
```
ctest --output-on-failure
```
Which runs all the tests?

(Otherwise there are missing `&& \` to get the failure if a test fails.)

test/scripts/test_clicReconstruction.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
```
andresailer, line 2: Is it possible to add to fail if any command fails, or if there is an undefined variable?
```suggestion
set -eu
```

.github/scripts/gmp_build_and_test.sh
fdplacido: agreed; before I was running one particular test, but now we just run all of them.

andresailer:
The first test is failing

andresailer (approved):

PR #21, opened 2020-10-30
Converter script changes

fdplacido:
BEGINRELEASENOTES

- steeringFileConverter
    - Add constants parsing of XML files
    - Add test case for multiple constants, lists and nested constants replacing

- Removed BOOST dependency

ENDRELEASENOTES


GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
+import re
+
+def parseConstants(constants):
+  again = True
```
andresailer, line 4: Need a docstring for this function. Including a few words of how it does its thing

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
...
+        formatted_list = ""
+        for val in value:
+          formatted_list += '{} '.format(val)
+        constants[key] = formatted_list
+      else:
+        if (value):
+          constants[key] = value % constants
+          again = (again) or (constants[key] != value)
+
+if __name__ == "__main__":
```
andresailer, line 21: This block can be dropped, the function call doesn't work like that because of the missing argument

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
+import re
+
+def parseConstants(constants):
+  again = True
+  while again:
+    again = False
+    for key,value in constants.items():
+      if type(value) is list:
```
andresailer, line 8:
```suggestion
      if isinstance(value, list):
```

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
+import re
+
+def parseConstants(constants):
+  again = True
+  while again:
+    again = False
+    for key,value in constants.items():
```
andresailer, line 7:
```suggestion
    for key, value in constants.items():
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
   return processors
 
 
+# Replace with constants if pattern found in value(s)
```
andresailer, line 59: docstrings like
```
def func(arg):
   """Do the thing.

   Need to use foo because bar.
   """
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+def replaceConstants(value, constants):
+  formatted_array = []
+  split_values = value.split()
+  for val in split_values:
+    captured_patterns = re.findall('\$\{\w*\}', val)
+    if len(captured_patterns) == 0:
```
andresailer, line 65:
```suggestion
    if not captured_patterns:
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+      formatted_array.append("\"{}\"".format(val))
+    elif len(captured_patterns) >= 1:
```
andresailer, line 67:
```suggestion
    elif captured_patterns:
```
although `else` would also just work

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+      val_format = re.sub(r'\$\{(\w*)\}', r'%(\1)s', val)
+      val_format = "\"{}\" % CONSTANTS".format(val_format)
```
andresailer, line 69:
```suggestion
      val_format = '"{}" % CONSTANTS'.format(val_format)
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+  constElements = tree.findall('constants/include')
+  if constElements:
+    print('WARNING: include statements found inside constants tag')
+
+  constElements = tree.findall('constants/constant')
+  for const in constElements:
+    constants[const.attrib.get('name')] = getValue(const)
+
+  for key,value in constants.items():
+    if value:
```
andresailer, line 88:
```suggestion
    if not value:
      continue
```
And dedent the rest of the loop block

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+
+  constElements = tree.findall('constants/include')
+  if constElements:
+    print('WARNING: include statements found inside constants tag')
+
+  constElements = tree.findall('constants/constant')
+  for const in constElements:
+    constants[const.attrib.get('name')] = getValue(const)
+
+  for key,value in constants.items():
```
andresailer, line 87:
```suggestion
  for key, value in constants.items():
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+            formatted_array.append("\"{}\"".format(val))
+          elif len(captured_patterns) >= 1:
+            val_format = re.sub(r'\$\{(\w*)\}', r'%(\1)s', val)
+            val_format = "\"{}\"".format(val_format)
+            formatted_array.append(val_format)
+        constants[key] = '[{}]'.format(", ".join(formatted_array))
+
+  lines.append("\nCONSTANTS = {")
+  for key in constants:
+    lines.append("\t\t'{}': {},".format(key, constants[key]))
```
andresailer, line 109: Is the rest of the code also using tabs for whitespace in the output file?
```python
' ' * len('CONSTANTS = {') + "{}: {},".format...
```
or something

GMPWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -157,8 +156,11 @@ std::shared_ptr<marlin::StringParameters> MarlinProcessorWrapper::parseParameter
       continue;
     }
 
-    parameterValues.push_back(parameterString);
+    auto split_parameter = gmp::util::split(parameterString);
+    for (auto& s : split_parameter) parameterValues.push_back(s);
```
andresailer, line 160:
```suggestion
    parameterValues.insert(parameterValues.end(), split_parameter.begin(), split_parameter.end());
```

GMPWrapper/src/util/gmpUtil.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
andresailer, line 1: Why not put this in this folder?
https://github.com/key4hep/GMP/tree/master/GMPWrapper/GMPWrapper

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
...
+      if type(value) is list:
+        for idx, val in enumerate(value):
+          value[idx] = val % constants
+          again = (again) or (value[idx] != val)
```
andresailer, line 11:
```suggestion
          again = again or (value[idx] != val)
```

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
...
+        formatted_list = ""
+        for val in value:
+          formatted_list += '{} '.format(val)
+        constants[key] = formatted_list
```
andresailer, line 15: These four lines are the same as
```suggestion
        constants[key] = " ".join(value)
```
right? (except for a trailing space)

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
...
+      else:
```
andresailer, line 16:
```suggestion
      elif value:
```
drop next if, dedent?

GMPWrapper/python/GMPWrapper/parseConstants.py
```diff
@@ -0,0 +1,22 @@
...
+        if (value):
+          constants[key] = value % constants
+          again = (again) or (constants[key] != value)
```
andresailer, line 19:
```suggestion
          again = again or (constants[key] != value)
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
 
 
+# Replace with constants if pattern found in value(s)
+def replaceConstants(value, constants):
+  formatted_array = []
+  split_values = value.split()
+  for val in split_values:
+    captured_patterns = re.findall('\$\{\w*\}', val)
+    if len(captured_patterns) == 0:
+      formatted_array.append("\"{}\"".format(val))
```
andresailer, line 66:
```suggestion
      formatted_array.append('"{}"'.format(val))
```

GMPWrapper/src/util/gmpUtil.h
```diff
@@ -0,0 +1,30 @@
...
+  const std::regex& re) 
+{
+  std::vector<std::string> container{
+      std::sregex_token_iterator(subject.begin(), subject.end(), re, -1), 
+      std::sregex_token_iterator()
+  };
+  return container;
+}
+
+// Split a string by space character
```
andresailer, line 22:
```suggestion
// Split a string by any whitespace characters
```

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -55,6 +56,64 @@ def getProcessors(tree):
...
+  constElements = tree.findall('constants/constant')
+  for const in constElements:
+    constants[const.attrib.get('name')] = getValue(const)
+
+  for key,value in constants.items():
+    if value:
+      formatted_array = []
+      split_values = value.split()
+      if len(split_values) == 1:
+        captured_patterns = re.findall('\$\{\w*\}', value)
```
andresailer, line 92: I think all the regex lines deserve a comment of what is being done

andresailer (changes requested):
Thanks!
A bit of cleanup and documentation for the python parts are still needed.

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
fdplacido: Isn't it the same though? it has to branch anyway, and only do something if there is a value.

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
fdplacido: No. The plan was to introduce this in the rest of the code, but for now I'll remove it as it makes it inconsistent.

GMPWrapper/src/components/MarlinProcessorWrapper.cpp
fdplacido: I would argue that `.push_back()` is more specific (and efficient) here, since I always want to put it at the end of the vector. The insert has to do more checks, and as a developer I have to actually pause and check where is the developer putting the element.

GMPWrapper/src/util/gmpUtil.h
fdplacido: I'm ok with that, but I still prefer to have it in a separate util folder.

GMPWrapper/scripts/convertMarlinSteeringToGaudi.py
andresailer: Yes it is the same, but easier to parse visually with shorter blocks and not these long blocks inside if and no else.

GMPWrapper/src/components/MarlinProcessorWrapper.cpp
andresailer: The insert has to check once for the range, the push back might cause allocation multiple times.
According to this https://stackoverflow.com/a/57521657 under their circumstances the insert is more efficient.

Add a comment for the developer.

GMPWrapper/src/util/gmpUtil.h
andresailer: Fine for me

GMPWrapper/GMPWrapper/MarlinProcessorWrapper.h
```diff
@@ -26,6 +26,8 @@
 
 #include <stack>
 
+#include <GMPWrapper/GMPWrapper/util/gmpUtil.h>
```
andresailer, line 29:
```suggestion
#include "util/gmpUtil.h"
```
Not sure if the change to quotes is needed, but the path should be like this, no?

fdplacido: totally

andresailer (approved):

PR #25, opened 2021-01-14
Plfernan gaudi35

fdplacido:
BEGINRELEASENOTES
- Adaptation and compilation for Gaudi v35

ENDRELEASENOTES


k4MarlinWrapper/CMakeLists.txt
```diff
@@ -19,40 +19,40 @@
...
-    ${LCIO_INCLUDE_DIRS}
-    ${Marlin_INCLUDE_DIRS}
-  )
+)
+
+target_include_directories(k4MarlinWrapperPlugins PUBLIC
+  k4MarlinWrapper/
+  ${LCIO_INCLUDE_DIRS}
+  ${Marlin_INCLUDE_DIRS}
+  /home/plfernan/workspace/json/install/include
```
andresailer, line 40: This shouldn't be here, and not be necessary locally either.

k4MarlinWrapper/CMakeLists.txt
```diff
@@ -19,40 +19,40 @@
...
+)
 
-target_compile_options(k4MarlinWrapperPlugins PUBLIC "-Wno-suggest-override")
-target_compile_options(MarlinWrapper PUBLIC "-Wno-suggest-override")
-gaudi_env(PREPEND LD_LIBRARY_PATH ${Gaudi_DIR}/lib/)
+target_include_directories(MarlinWrapper PUBLIC
+  k4MarlinWrapper/
+  ${Marlin_INCLUDE_DIRS}
+  ${LCIO_INCLUDE_DIRS}
+  /home/plfernan/workspace/json/install/include
```
andresailer, line 57: This shouldn't be here, and not be necessary locally either.

CMakeLists.txt
```diff
@@ -34,22 +34,25 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
 set(CMAKE_CXX_FLAGS_RELEASE "-O3 -DNDEBUG")
 set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g")
 
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "")
+
+if(NOT CMAKE_CXX_STANDARD MATCHES "14|17")
```
vvolkl, line 39: Does it make sense to already add
```suggestion
if(NOT CMAKE_CXX_STANDARD MATCHES "14|17|20")
```
?

cmake/gaudi_binary_tag.cmake
```diff
@@ -4,4 +4,4 @@
 if( NOT DEFINED ENV{LCG_VERSION})
```
andresailer, line 4: Is this still need now?

fdplacido: Not anymore, so I'll remove the file.

CMakeLists.txt
fdplacido: We are not using any C++20 stuff. I'm not sure about this one:
- By putting it there, it may seem like a C++20 compatible compiler is needed.
- On the other hand, it is optional as indicated, so we are welcoming to use a C++20 feature if someone wants to.

vvolkl: My consideration was that you might have a global setting of the cxx standard  (spack allows you to set the cxxstd as a global variant (see https://github.com/key4hep/key4hep-spack/issues/149)), and setting it to 20 would then result in cmake error here, even though it would build fine.

fdplacido: Yeah, that makes sense, I'll add it.

vvolkl (approved):

andresailer (approved):

PR #26, opened 2021-02-05
EDM4hep <--> LCIO converters

fdplacido:
BEGINRELEASENOTES
- Implement EDM4hep to LCIO converter on the fly
- Add LCIO to EDM4hep converter on the fly through k4LCIOReader
- Gaudi Tools to be added to Algorithms on demand on options file
- Added testing for EDM4hep -> LCIO conversion and EDM4hep -> LCIO -> EDM4hep conversion, creating the collections artificially, in place.
- Changed testing to get input files from HTTP key4hep
- Added documentation on usage of converters. Updated documentation.

ENDRELEASENOTES


test/gaudi_opts/test_edm4hep2lcio.py
```diff
@@ -0,0 +1,48 @@
...
+
+procA = MarlinProcessorWrapper("TestProcessor")
+procA.OutputLevel = DEBUG
+procA.ProcessorType = "AIDAProcessor"
+procA.Parameters = ["FileName", "histograms", END_TAG,
+                    "FileType", "root", END_TAG,
+                    "Compress", "1", END_TAG,
+                    "Verbosity", "DEBUG", END_TAG
+                    ]
+procA.Conversion = ["edm4hep::ReconstructedParticleCollection", "ReconstructedParticles", "VertexBarrelCollection",
```
andresailer, line 34: Different collection names

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
```
vvolkl, line 1: We discussed `pragma once` and said to prefer old-style include guards, no?

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -121,6 +121,31 @@ StatusCode MarlinProcessorWrapper::loadProcessorLibraries() const {
...
+  // Check if there is an event already, if not convert the input EDM4hep collection
+  info() << "Looking for event in " << m_processor->name() << " processor." << endmsg;
+  DataObject* pObject = nullptr;
+  StatusCode  sc      = eventSvc()->retrieveObject("/Event/LCEvent", pObject);
+  if (sc.isFailure()) {
+    if (parameters.size() != 0) {
+      info() << "No event found. Converting EDM4hep to LCIO. " << m_processor->name() << endmsg;
+      // Convert parameters
+      m_conversionTool = tool<IEDM4hep2LcioTool>("EDM4hep2LcioTool/Converter");
+      StatusCode sc1 =  m_conversionTool->convertCollections(parameters, lcio_event);
```
andresailer, line 138: Check if the collection exists in the event, if not convert it.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -0,0 +1,241 @@
...
+    error() << " Error processing conversion parameters. 3 arguments per collection expected. " << endmsg;
+    return StatusCode::FAILURE;
+  }
+
+  for (int i = 0; i < parameters.size(); i=i+3) {
+    convert(parameters[i], parameters[i+1]);
+  }
+
+  // Populate LCIO event with EDM4hep info
+  auto* tracks = new lcio::LCCollectionVec(lcio::LCIO::TRACK);
```
andresailer, line 211: Need to loop over the collections to convert, and then create those (unless they already exist in the LCIO event)

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -0,0 +1,241 @@
...
+  lcio_event->addCollection(tracks, lcio::LCIO::TRACK);
+
+  // Add converted reconstructed particles to LCIO event
+  for (auto& rp : lcio_rec_particles_vec) {
+    recops->addElement(rp.first);
+  }
+  lcio_event->addCollection(recops, lcio::LCIO::RECONSTRUCTEDPARTICLE);
+
+  // Register LCIO event in TES
+  info() << "Registering converted EDM4hep to LCIO event in TES" << endmsg;
```
andresailer, line 232: Probably not neeed, can just assume that the LCIO event is in the data store (this must be handled somewhere else)

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
fdplacido: done.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -0,0 +1,276 @@
...
+    const edm4hep::Track edm_tr = (*tracks_coll)[i];
+
+    auto* lcio_tr = new lcio::TrackImpl();
+
+    lcio_tr->setTypeBit( edm_tr.getType() );
+    lcio_tr->setChi2( edm_tr.getChi2() );
+    lcio_tr->setNdf( edm_tr.getNdf() );
+    lcio_tr->setdEdx( edm_tr.getDEdx() );
+    lcio_tr->setdEdxError( edm_tr.getDEdxError() );
+    lcio_tr->setRadiusOfInnermostHit( edm_tr.getRadiusOfInnermostHit() );
```
andresailer, line 31:
```suggestion
    lcio_tr->setRadiusOfInnermostHit( edm_tr.getRadiusOfInnermostHit() );
    // FIXME, correctly assign hitnumbers
    lcio_tr->subdetectorHitNumbers().resize(50);
    for(int i =0;i<50;++i) {
      lcio_tr->subdetectorHitNumbers()[i] = 0 ; 
    }
```

andresailer: Maybe the conversion from edm4hep was forgotten?
https://github.com/key4hep/EDM4hep/blob/1bd0b5db00ec8659c53041e07784c34f8cc01add/edm4hep.yaml#L325

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -0,0 +1,272 @@
...
+    const podio::RelationRange<edm4hep::TrackState> edm_track_states = edm_tr.getTrackStates();
+    for (const auto& tr_state : edm_track_states) {
+
+      std::array<float, 15> cov = {}; // zero-init
+      cov[0]  = tr_state.covMatrix[0];
+      cov[2]  = tr_state.covMatrix[5];
+      cov[5]  = tr_state.covMatrix[9];
+      cov[9]  = tr_state.covMatrix[12];
+      cov[14] = tr_state.covMatrix[14];
+      std::array<float, 3> refP = {
```
clementhelsens, line 48: Why do we give only the diagonal terms? The fact that in k4DelphesSim we store only them is a something we are aware and will fix

andresailer: Because that is what was present before in your converter.
Also we need to sort this out https://github.com/key4hep/k4LCIOReader/issues/8
and this
https://github.com/key4hep/k4SimDelphes/blob/8561d3fc39b2accc99c62a561090421fad3f1e8e/converter/src/DelphesEDM4HepConverter.cc#L488-L493

andresailer:
Adding conversion from LCIO back to EDM4hep

- [ ] allow configuring which collections to transform. Can use the Conversion parameter and have 'lcio::...' as the first entry in the field to indicate the conversion is going the other way. Or adding a second parameter, in which case maybe `Conversion` should be renamed and `edm4hep::` or `lcio::` wouldn't be needed.
- [ ] Calls to the k4LCIOReader conversion functionality
- [ ] Storing the output edm4hep collections in the podio/edm4hep file.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
clementhelsens: ok, I see thanks. With changing upper to lower it will make it more clear than placing individual element of the matrix

fdplacido: I implemented it this way as I took it from the python example, but I will store all the elements transposed.

andresailer: Following https://github.com/key4hep/EDM4hep/issues/107 there should be no transposition

fdplacido: I changed it to store the full matrix, no transpose. This produces some (many) `helix variance is negative: -X`, but that will change once the changes from EDM4hep are included.

fdplacido:
There is an initial version of the converter back from LCIO to EDM4hep using k4LCIOReader. Note that this PR: https://github.com/key4hep/k4LCIOReader/pull/9 needs to be merged for the converter to be able to access the converter.

test/gaudi_opts/test_edm4hep2lcio.py
```diff
@@ -0,0 +1,77 @@
+from Gaudi.Configuration import *
+
+from Configurables import k4DataSvc, MarlinProcessorWrapper, EDM4hep2LcioTool, k4LCIOReaderWrapper
+
+algList = []
+
+END_TAG = "END_TAG"
+
+# theFile= 'edminput.root'
+theFile = '/eos/experiment/fcc/ee/generation/DelphesEvents/fcc_tmp/p8_ee_Ztautau_ecm91_EvtGen_Tau2MuGamma/events_001720714.root'
```
andresailer, line 10: @clementhelsens Do you have  a file created with the latest delphes with the full and "transposed" covariance matrix"?

andresailer: from @vvolkl: put the file in an http accessible place

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -219,21 +223,43 @@ StatusCode MarlinProcessorWrapper::initialize() {
...
+
+  lcio::LCEventImpl* the_event = nullptr;
+
   if (sc.isFailure()) {
-    error() << "Failed to retrieve the LCEvent " << endmsg;
-    return sc;
+    the_event = new lcio::LCEventImpl();
+    // Register empty event
+    debug() << "Registering conversion EDM4hep to LCIO event in TES" << endmsg;
+    auto pO = std::make_unique<LCEventWrapper>(the_event);
```
andresailer, line 238: I think we still have to add the "ownership" or "cleanup" parameter to the LCEventWrapper to indicate of the_event should be cleaned in the end or not. Depending on if the event came from the LCReader or is created like this here.

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -219,21 +223,43 @@ StatusCode MarlinProcessorWrapper::initialize() {
...
+    StatusCode reg_sc = evtSvc()->registerObject("/Event/LCEvent", pO.release());
+    if (reg_sc.isFailure()) {
+      error() << "Failed to store the EDM4hep to LCIO event" << endmsg;
+      return reg_sc;
+    }
+  } else {
+    debug() << "LCEvent retrieved successfully" << endmsg;
+    the_event =
+      dynamic_cast<IMPL::LCEventImpl*>(static_cast<LCEventWrapper*>(pObject)->getEvent());
```
andresailer, line 247: Change the type of `m_event` in LCEventWrapper?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
fdplacido: fixed

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -0,0 +1,190 @@
...
+#include "edm4hep/TPCHitCollection.h"
+#include "edm4hep/Cluster.h"
+#include "edm4hep/ClusterCollection.h"
+#include "edm4hep/Vertex.h"
+#include "edm4hep/VertexCollection.h"
+#include "edm4hep/MCParticle.h"
+#include "edm4hep/MCParticleCollection.h"
+
+// LCIO
+#include <LCEventWrapper.h>
```
andresailer, line 38: Better to change the include directories so this has to be prefixed by the folder
```suggestion
#include <k4MarlinWrapper/LCEventWrapper.h>
```

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -0,0 +1,190 @@
...
+
+#include "LCIOSTLTypes.h"
+
+#include <vector>
+#include <string>
+#include <bitset>
+#include <algorithm>
+
+// Interface
+#include "IEDMConverter.h"
```
andresailer, line 65: also for this one

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
fdplacido: I would suggest renaming `k4MarlinWrapper/k4MarlinWrapper` to `k4MarlinWrapper/include` and keep the includes as they are: putting the files names from inside the include folder. That way we have the `src` and `include` folders inside the k4MarlinWrapper folder and avoid having a triple nested k4MarlinWrapper path.

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
fdplacido: This I would change to `converters/IEDMConverters.h`.

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
andresailer: I would like to have "k4MarlinWrapper" before, because if the header files are installed they don't end up in a global "include" folder but in "include/k4marlinwrapper" when doing a"view" type thing.

PR #28, opened 2021-06-11
fix check of bounds on parameters sorting

fdplacido:
BEGINRELEASENOTES
- Fix checks on second parameter when sorting them.

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
@@ -23,35 +23,47 @@ void optimizeOrderParams(
     // if target before dependency swap all related elements. Multiply indexes by 3 to get correct index from view.
     auto swap_if_before = [&params_view, &params](const std::string& target, const std::string& dependency)
     {
-      auto target_found_it = std::find(params_view.begin(), params_view.end(), target);
-      if (target_found_it != params_view.end()) {
+      const int maxidx = 5;
+      int gidx = 0;
+      // Look for instances of target
+      for (int t_idx=0; t_idx<params_view.size(); ++t_idx) {
+        if (params_view[t_idx] == target) {
+          // Look for instances of dependency
+          for (int d_idx=0; d_idx<params_view.size(); ++d_idx) {
+            if (params_view[d_idx] == dependency) {
+              // If target is before the depenency swap them
+              if (t_idx < d_idx) {
+                const auto real_t_idx = 3 * t_idx;
+                const auto real_d_idx = 3 * d_idx;
 
-        auto target_index = 3 * std::distance(params_view.begin(), target_found_it);
-        auto dependency_found_it = std::find(params_view.begin(), params_view.end(), dependency);
+                std::swap(params[real_t_idx], params[real_d_idx]);
+                std::swap(params[real_t_idx+1], params[real_d_idx+1]);
+                std::swap(params[real_t_idx+2], params[real_d_idx+2]);
+                // swap also the view
+                std::swap(params_view[t_idx], params_view[d_idx]);
+                // reset target index until everything is sorted
+                t_idx = 0;
 
-        if (dependency_found_it != params_view.end()) {
-
-          auto dependency_index = 3 * std::distance(params_view.begin(), dependency_found_it);
-
-          if (target_index < dependency_index) {
-            std::swap(params[target_index], params[dependency_index]);
-            std::swap(params[target_index+1], params[dependency_index+1]);
-            std::swap(params[target_index+2], params[dependency_index+2]);
+                break;
+              }
+            }
           }
         }
       }
     };
 
     // A depends on B converted before to not need FillMissingCollections()
-    swap_if_before("Track", "TrackerHit");
-    swap_if_before("SimTrackerHit", "MCParticle");
-    swap_if_before("SimCalorimeterHit", "MCParticle");
-    swap_if_before("Cluster", "CalorimeterHit");
-    // swap_if_before("Vertex", "ReconstructedParticle");
-    swap_if_before("ReconstructedParticle", "Track");
-    // This generates a cycle, but that's fine
-    // swap_if_before("ReconstructedParticle", "Vertex");
-    swap_if_before("ReconstructedParticle", "Cluster");
+    for (int i=0; i < 2; ++i) {
+      swap_if_before("Track", "TrackerHit");
+      swap_if_before("SimTrackerHit", "MCParticle");
+      swap_if_before("SimCalorimeterHit", "MCParticle");
+      swap_if_before("Cluster", "CalorimeterHit");
+      // swap_if_before("Vertex", "ReconstructedParticle");
+      swap_if_before("ReconstructedParticle", "Track");
+      // This generates a cycle, but that's fine
+      // swap_if_before("ReconstructedParticle", "Vertex");
+      swap_if_before("ReconstructedParticle", "Cluster");
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -1162,14 +1162,16 @@ void EDM4hep2LcioTool::optimizeOrderParams()
...
         auto track_index = std::distance(m_edm2lcio_params.begin(), track_found_it);
-
         auto trackerhit_found_it = std::find(m_edm2lcio_params.begin(), m_edm2lcio_params.end(), dependency);
-        auto trackerhit_index = std::distance(m_edm2lcio_params.begin(), trackerhit_found_it);
 
-        if (track_index < trackerhit_index) {
-          std::swap(m_edm2lcio_params[track_index], m_edm2lcio_params[trackerhit_index]);
-          std::swap(m_edm2lcio_params[track_index+1], m_edm2lcio_params[trackerhit_index+1]);
-          std::swap(m_edm2lcio_params[track_index+2], m_edm2lcio_params[trackerhit_index+2]);
+        if (trackerhit_found_it != m_edm2lcio_params.end()) {
```
andresailer, line 1167: this way less indentation changes are needed below.
```suggestion
        if (trackerhit_found_it == m_edm2lcio_params.end()) {
          continue;
        }
```

vvolkl:
Could you add a CMake option  `USE_EXTERNAL_CATCH2` like in k4simdelphes

https://github.com/key4hep/k4SimDelphes/blob/main/tests/unittests/CMakeLists.txt#L2
https://github.com/key4hep/k4SimDelphes/blob/main/CMakeLists.txt#L111

? This will let us build a static library for catch2 once  instead of rebuilding it for every repository.

fdplacido:
> Could you add a CMake option `USE_EXTERNAL_CATCH2` like in k4simdelphes
> 
> https://github.com/key4hep/k4SimDelphes/blob/main/tests/unittests/CMakeLists.txt#L2
> https://github.com/key4hep/k4SimDelphes/blob/main/CMakeLists.txt#L111
> 
> ? This will let us build a static library for catch2 once instead of rebuilding it for every repository.

I implemented it, thanks!

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,60 @@
...
+      params_view.emplace_back(params[i]);
+    }
+
+    // Can't run in parallel
+    // Dependency should be converted before the target, so swap them (and their names) if out of order.
+    // search for the target and its dependency string, get indexes,
+    // if target before dependency swap all related elements. Multiply indexes by 3 to get correct index from view.
+    auto swap_if_before = [&params_view, &params](const std::string& target, const std::string& dependency)
+    {
+      auto target_found_it = std::find(params_view.begin(), params_view.end(), target);
```
andresailer, line 26: you need to find all the targets, as there can be more than one

fdplacido: I changed the algorithm to take into account multiple repeated types, so it now searches for all targets, dependencies, etc.
I added a unit test with that case.

fdplacido:
I changed the algorithm to take into account multiple repeated types, so it now searches for all targets, dependencies, etc.
I added a unit test with that case.

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,60 @@
...
+
+        auto target_index = 3 * std::distance(params_view.begin(), target_found_it);
+        auto dependency_found_it = std::find(params_view.begin(), params_view.end(), dependency);
+
+        if (dependency_found_it != params_view.end()) {
+
+          auto dependency_index = 3 * std::distance(params_view.begin(), dependency_found_it);
+
+          if (target_index < dependency_index) {
+            std::swap(params[target_index], params[dependency_index]);
```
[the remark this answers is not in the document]
fdplacido: I didn't know about this. It needs to work with iterators which in this case would add more lines. I think for 3 cases its still simpler this way.

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,72 @@
...
+      params_view.emplace_back(params[i]);
+    }
+
+    // Can't run in parallel
+    // Dependency should be converted before the target, so swap them (and their names) if out of order.
+    // search for the target and its dependency string, get indexes,
+    // if target before dependency swap all related elements. Multiply indexes by 3 to get correct index from view.
+    auto swap_if_before = [&params_view, &params](const std::string& target, const std::string& dependency)
+    {
+      const int maxidx = 5;
```
andresailer, line 26: unused?

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,72 @@
...
+      int gidx = 0;
```
andresailer, line 27: unused?

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,72 @@
...
+                break;
+              }
+            }
+          }
+        }
+      }
+    };
+
+    // A depends on B converted before to not need FillMissingCollections()
+    for (int i=0; i < 2; ++i) {
```
andresailer, line 56: why this loop?

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,72 @@
...
+    {
+      const int maxidx = 5;
+      int gidx = 0;
+      // Look for instances of target
+      for (int t_idx=0; t_idx<params_view.size(); ++t_idx) {
+        if (params_view[t_idx] == target) {
+          // Look for instances of dependency
+          for (int d_idx=0; d_idx<params_view.size(); ++d_idx) {
+            if (params_view[d_idx] == dependency) {
+              // If target is before the depenency swap them
```
andresailer, line 34:
```suggestion
              // If target is before the dependency swap them
```

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
fdplacido: With just 1 loop, some collections will go "up" again in the list and some targets will stay before the dependency. Going over them again fixes that.

k4MarlinWrapper/k4MarlinWrapper/util/EDM4hep2LcioUtils.h
```diff
@@ -0,0 +1,71 @@
...
+namespace k4MW::util {
+
+// Reorder parameters based on dependencies between collections
+// This avoids (partly) the need for FillMissingCollections
+template <typename T>
+void optimizeOrderParams(
+  T& params)
+{
+
+  if (params.size() % 3 == 0) {
```
andresailer, line 13: Do you really need this check? Is this not checked much earlier already and then issuing a fatal error?

fdplacido: It is indeed, this was an early attempt to make the method "more independent" from the only place that it can be really called. But I think that's a branch we can remove.

andresailer: yes, either remove this branch or throw an exception if this isn't fulfilled?

fdplacido: I removed to not keep two checks.

andresailer (approved):

Issue #36, opened 2021-08-05
"EOF" is not handled gracefully by LcioEvent and the MarlinProcessorWrapper combination


tmadlener:
Currently the `LcioEventAlgo` does not handle the following two cases gracefully:
- Specifying to run over a number of events in the options file (or via `--num-events`) that is larger than the number of events in the input file.
- Specifying to run over all events via `--num-events=-1` (or `EvtMax = -1` in the options file)

The underlying reason for both problems is probably the same. When the input file is exhausted `LcioEventAlgo::execute` returns `StatusCode::FAILURE` but the event processing will continue. Since the `LcioEventAlgo` didn't register an `LCEvent` with the event service the `MarlinProcessorWrapper` will create a new but empty `LCEvent` and register that with the event service. That empty event will then break at some point in the chain of wrapped Marlin processors, where the first processor does not gracefully handle missing collections in the event.

PR #42, opened 2021-09-13
Add sequence stop on not finding more events to process

fdplacido:
BEGINRELEASENOTES
- When indicating to process more events than available, algorithm will create empty event and fail by trying to continue.
- If no more events to read are found, the processing will stop
ENDRELEASENOTES


fdplacido:
Fixes https://github.com/key4hep/k4MarlinWrapper/issues/36

tmadlener (commented):
Hi @fdplacido,
Thanks for taking care of this. Looks good to me. 

I have two minor questions that only concern the tests: Wouldn't it be possible to run the tests via `k4run [...] --num-events=[number of events]`? In that way only one options file and bash script would be necessary to cover both currently present tests. Also I think it would be nice to add a comparison of the number of events in the input file and the output file via `lcio_event_counter` (just to be really sure that there is no empty event).

fdplacido:
I changed it to use `k4run` as you suggested. For the number of events comparison: I agree, that is a good test to have, so I'll create a simple options file that creates some output to be able to compare.

fdplacido:
I added a test to check that using `lcio_event_counter`, thanks for the suggestion @tmadlener

test/scripts/same_num_io.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ ! -d $k4MarlinWrapper_tests_DIR/inputFiles/ ]; then
+  mkdir $k4MarlinWrapper_tests_DIR/inputFiles
+fi
+
+if [ ! -f $k4MarlinWrapper_tests_DIR/inputFiles/muons.slcio ]; then
+  wget https://github.com/AIDASoft/DD4hep/raw/master/DDTest/inputFiles/muons.slcio -P $k4MarlinWrapper_tests_DIR/inputFiles/
+fi
+
+../run k4run $k4MarlinWrapper_tests_DIR/gaudi_opts/same_num_io.py
```
tmadlener, line 12: Could this be run with `--num-events` set to `-1` or a number greater than the input events, otherwise we are basically just testing the `EvtMax` handling of Gaudi itself, instead of testing that the k4MarlinWrapper handles EOF gracefully.

I think the original tests with `-1` and `15` as arguments to `--num-events` are exactly what we need, I would just put the check for the same number of events in the inputs and outputs to those tests, using a gaudi options file that also produces an `slcio` output file (I think the one that is used in those tests, doesn't produce an output(?)).

test/scripts/same_num_io.sh
```diff
@@ -0,0 +1,21 @@
...
+
+input_num_events=$(lcio_event_counter $k4MarlinWrapper_tests_DIR/inputFiles/muons.slcio)
+output_num_events=$(lcio_event_counter Output_DST.slcio)
+
+if [ "$input_num_events" == "$output_num_events" ] || [ "$input_num_events" == 10 ]; then
```
tmadlener, line 17:
```suggestion
if [ "$input_num_events" == "$output_num_events" ]; then
```

Otherwise we would not get an error even if the two numbers are different as long as the number of input events is 10. As said above, what we should be testing is that cases where we want all events, or more events than are present are handled as expected.

tmadlener (changes requested):

test/scripts/same_num_io.sh
fdplacido: my bad, I didn't realize that this way the test was a bit useles. Thanks!

fdplacido:
I updated as you suggested, checking now for correct number of output events with `-1`, and testing for any number of events.

tmadlener (approved):

Issue #35, opened 2021-08-05
Differences in XML parsing between Marlin and the converter script


tmadlener:
It looks like there are some differences between the xml that is considered valid by `Marlin` and the xml that is considered valid by the `convertMarlinSteeringToGaudi.py` script. Here are two examples that I just encountered.

### Treating `&` as escape character or not
Considering the following [line from a steering file](https://github.com/iLCSoft/ILDConfig/blob/6bf49e0b7d36074ed0987d03b8721ba1c16e102d/StandardConfig/production/MarlinStdRecoMiniDST.xml#L330)
```xml
    <parameter name="FlavorTag.CategoryDefinition2" type="string">nvtx==1&&nvtxall==1</parameter>
```
`Marlin` seems to parse this just fine and work without any problems, but trying to convert the steering file results in a parsing error:
```console
Exception when getting trees: ParseError('not well-formed (invalid token): line 330, column 72') 
```
It looks like `Marlin` is a bit more permissive here, since `&` is an escape character in XML.

### Automatically replacing environment variables
From the same steering file, consider [this line](https://github.com/iLCSoft/ILDConfig/blob/6bf49e0b7d36074ed0987d03b8721ba1c16e102d/StandardConfig/production/MarlinStdRecoMiniDST.xml#L45)
```xml
  <constant name="lcgeo_DIR" value=$lcgeo_DIR/>
```
Here `Marlin` again happily replaces `$lcgeo_DIR` with the value of that environment variable but the converter script runs into a parsing error:
```console
Exception when getting trees: ParseError('not well-formed (invalid token): line 45, column 35')
```
I am not entirely sure if this replacement happens automatically within the xml parser used by `Marlin` or if this is done deliberately.

-----
This seems to be a continuation of #18 but I am not sure if a fix exists for these new ones here. They can be fixed by changing the original steering file and/or by parsing it with `Marlin` first to do some of the replacements and producing a parsed steering file from the original one and then passing the parsed one to the converter.

As "quick fixes" the following can be done:
- replace all `&&` with `&amp;&amp` in the original steering file
- Do the replacement of the environment variables yourself or use the `global.OutputSteeringFile` parameter and `Marlin -n --global.OutputSteeringFile=parsed.xml <other-constants-to-replace> steering-file.xml`. (Note that you might have to introduce the `OutputSteeringFile` parameter in the `global` section first to be able to override it)

andresailer:
[...]

PR #43, opened 2021-09-13
fix ampersand & character failing on xml to py conversion

fdplacido:
BEGINRELEASENOTES
- convertMarlinSteeringToGaudi: When an escape XML character was found (&), reading the XML Tree would fail. The character is nowescaped, so the marlin XML file is processed correctly

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
+++ b/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
@@ -248,7 +248,7 @@ def generateGaudiSteering(tree):
 
 
 def escapeIllegalChars(file_str):
-  return file_str.replace("&", "&amp;")
+  return file_str.replace("&&", "&amp;&amp;")
 
 
 def run():
--- a/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
+++ b/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
@@ -6,14 +6,7 @@ from copy import deepcopy
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


fdplacido:
Fixes https://github.com/key4hep/k4MarlinWrapper/issues/35

k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
```diff
@@ -254,15 +247,22 @@ def generateGaudiSteering(tree):
   return lines
 
 
+def escapeIllegalChars(file_str):
+  return file_str.replace("&", "&amp;")
```
tmadlener, line 251:
```suggestion
  return file_str.replace("&&", "&amp;&amp;")
```
Maybe we should consider only escaping double ampersands, as single ampersands could still be valid use-cases of xml escaping(?). At least #35 only tripped over the double ampersands, so it might be enough.

fdplacido: Indeed, if some valid escaping goes into the XML like `&amp;` or `&quot;`, that is going to mess things up. So if we can just cover the case with double ampersands it will be much simpler.

andresailer:
What does the parameter value look like in the python result? `&&` or  `&amp;&amp;` ?

tmadlener:
> What does the parameter value look like in the python result? `&&` or `&amp;&amp;` ?

It should be `&&`.

fdplacido:
It is `&&` in the converted python file

andresailer:
Ok, good. :+1:

andresailer (approved):

Issue #45, opened 2021-09-27
Doctest in Github actions fails


fdplacido:
Fails after installing:

```sh
Successfully installed jupytext-1.13.0 markdown-it-py-1.1.0 mdit-py-plugins-0.2.8 typing-extensions-3.10.0.2
Defaulting to user installation because normal site-packages is not writeable
Requirement already up-to-date: jupyter in /cvmfs/sw.hsf.org/spackages2/py-jupyter/1.0.0/x86_64-centos7-gcc8.3.0-opt/grqxrxlo4jzt5sbztxhurtabmvx5qaxu/lib/python3.7/site-packages (1.0.0)
Traceback (most recent call last):
```

```sh
"[Errno 28] No space left on device: '/tmp/tmpefxs4rc0_kernels/python3'")]
Error: Process completed with exit code 1.
```

andresailer:
Tests are still failing, but now because some python packages are missing

https://github.com/key4hep/k4MarlinWrapper/runs/4296865385?check_suite_focus=true#step:7:35
RuntimeError: An error occurred while executing the notebook. Please make sure that 'nbconvert' and 'ipykernel' are installed.

vvolkl:
This was working again in the meantime, but there were some changes requested to the recipe when I pushed it upstream recently, I'm checking now if those are causing this problem.

andresailer:
Solved in #57 , need to enable the doctest against the release eventually again.

PR #49, opened 2021-10-14
[ci] use jupytext from stack

vvolkl:
Jupytext should be set up in the nightlies now, this may already fix the problems running the doctest on github actions

BEGINRELEASENOTES
- [ci] use jupytext from stack
ENDRELEASENOTES


vvolkl:
Fixes #45

PR #50, opened 2021-10-22
LCIO 2 EDM4hep CellID string conversion

fdplacido:
BEGINRELEASENOTES
- Convert LCIO cellID string to metadata when using LCIO2EDM4hep conversion

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
+++ b/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
@@ -51,26 +51,6 @@ void Lcio2EDM4hepTool::convertRegister(const std::string& edm_name, const std::s
     return;
   }
 
-  // Convert Metadata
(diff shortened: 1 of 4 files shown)


k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -34,144 +34,199 @@ StatusCode Lcio2EDM4hepTool::finalize() {
...
     }
   }
   return false;
 }
 
+// **********************************
+// Convert all collections indicated in Tool parameters
+// Some collections implicitly convert associated collections, as for
+// key4hep/k4LCIOReader
+// Convert assoicated collections from LCRelation for existing EDM4hep relations
```
andresailer, line 136:
```suggestion
// Convert associated collections from LCRelation for existing EDM4hep relations
```

PR #51, opened 2021-11-11
[cmake] fix wrong include for targets

vvolkl:
Just noticed that this causes some nonsensical includes like `-I/tmp/arch/spack-stage/spack-stage-k4marlinwrapper-0.4-kahcs4rkiy7umpchldefwqigzdvy6fst/spack-src/test/EDM4HEP::edm4hep`.

Linking the target should be enough.

BEGINRELEASENOTES
- [cmake] fix wrong include for targets

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,16 +48,12 @@ include(cmake/compiler_output.cmake)
 
 include(GNUInstallDirs)
 
-find_package(ROOT 6.0 REQUIRED)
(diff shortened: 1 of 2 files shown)


fdplacido:
Thanks for bringing this up. I went ahead and did a general clean-up of the targets.

fdplacido (approved):

fdplacido:
A rebase is needed to merge this.

PR #55, opened 2021-11-24
Add test for clicReconstruction with EDM4hep input

fdplacido:
BEGINRELEASENOTES
- Test clicReconstruction sequence with EDM4hep input
- Use converters for both EDM4hep <--> LCIO conversion in all relevant algorithms
- Output reconstruction in EDM4hep format

ENDRELEASENOTES


andresailer:
We also should add this to the documentation

fdplacido:
> We also should add this to the documentation

I updated the docs on how to run with EDM4hep input.

andresailer (approved):

PR #67, opened 2022-02-10
add FCCee reco configuration

andresailer:
BEGINRELEASENOTES
- GaudiOpts: add example for FCCee_o1_v04 reconstruction based on https://github.com/iLCSoft/CLICPerformance/blob/master/fcceeConfig/fccReconstruction.xml including conversions from EDM4hep to LCIO at the start and to EDM4hep at the end
- GaudiOpts: CLIC Reco: add some missing calo hit Contributions collections, BeamCal Collection
ENDRELEASENOTES


fdplacido:
We have the new test file, but it isn't running as a test. 
I like to have all the options files in the same place, so to me this is fine.

andresailer:
I should add running the simulation and reco for FCCee to the documentation, and then it would run as a test inside there

test/gaudi_opts/clicRec_e4h_input.py
```diff
@@ -36,23 +36,28 @@
   'VertexEndcapCollection',
   'InnerTrackerBarrelCollection',
   'OuterTrackerBarrelCollection',
+  'InnerTrackerEndcapCollection',
+  'OuterTrackerEndcapCollection',
   'ECalEndcapCollection',
   'ECalEndcapCollectionContributions',
   'ECalBarrelCollection',
+  'ECalBarrelCollectionContributions'
```
andresailer, line 44:
```suggestion
  'ECalBarrelCollectionContributions',
```

ARGH

PR #73, opened 2022-05-12
fix cmake includes

fdplacido:
BEGINRELEASENOTES
- Fix include dirs dependencies in cmake

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


tmadlener (approved):

tmadlener:
Not sure if the test failure is related to the changes here?

vvolkl:
Looks like the test failures are related to the LD_LIBRARY_PATH that is no longer set in the key4hep-stack. I'm working on refactoring the rpath settings, but you could try adding this: https://github.com/HEP-FCC/k4Gen/blob/main/CMakeLists.txt#L6 explicitly

PR #85, opened 2022-06-24
Add options file for CED event display via wrapper

tmadlener:
BEGINRELEASENOTES
- Add a gaudi options file that can be used to run the event display using the CLIC detector via the Marlin wrapper.

ENDRELEASENOTES

@fdplacido I am not entirely sure where to best put this. Having such a large file purely in the documentation as an inline code block seems a bit excessive to me. If we had it somewhere in the repository, I could link from the documentation to the complete file and focus the documentation on the essential parts. What do you think? Also if there is a better suited place for this let me know.

### DIFF
--- a/k4MarlinWrapper/examples/event_display.py
+++ b/k4MarlinWrapper/examples/event_display.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+
+from Gaudi.Configuration import *
+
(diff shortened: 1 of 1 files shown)


fdplacido:
I think the file is fine here. The documentation file should just link to it, maybe inlining some relevant lines.

PR #92, opened 2022-11-16
Make `isFirstEvent` work as expected in wrapped processors

tmadlener:
BEGINRELEASENOTES
- Fix an issue where `isFirstEvent` always returns `false` in wrapped processors by calling `setFirstEvent` after processing the first event.

ENDRELEASENOTES

This was discovered by Jenny and if not fixed, would mean that processors relying on this would need quite a bit of work, see e.g. https://github.com/iLCSoft/MarlinKinfitProcessors/pull/19

- [x] Needs iLCSoft/Marlin#47

### DIFF
--- a/k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
+++ b/k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
@@ -249,7 +249,10 @@ StatusCode MarlinProcessorWrapper::execute() {
     } else {
       m_processor->processEvent(the_event);
     }
+
+    m_processor->setFirstEvent(false, marlin::ExternalProcessorMgrAccessor{this});
   }
+
   // Handle exceptions that may come from Marlin
   catch (marlin::SkipEventException& e) {
     // Store flag to prevent the rest of the event from processing
--- a/doc/starterkit/k4MarlinWrapperCLIC/howtoMultithread.md
+++ b/doc/starterkit/k4MarlinWrapperCLIC/howtoMultithread.md
@@ -12,6 +12,7 @@
 - Writing EDM4hep events with `PodioOutput()`
 - Writing EDM4hep events with `MarlinProcessorWrapper` of type `LCIOOutputProcessor`
 - Running non-thread algorithms/processors in parallel
+- Running wrapped Marlin processors that make use of the `isFirstEvent` method in their `processEvent` method to do some setup only in the first event. There is no way to make this thread safe. If you want your processor to be usable on in a multi threaded environment, you have to move this setup to `init`.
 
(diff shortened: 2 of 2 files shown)


k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -249,7 +249,10 @@ StatusCode MarlinProcessorWrapper::execute() {
     } else {
       m_processor->processEvent(the_event);
     }
+
+    m_processor->setFirstEvent(false, this);
```
andresailer, line 253: This isn't thread-safe though.

andresailer: Especially as this comes after processEvent, so this means that not only are things not reproducible because of the randomseed, but sequential is the only way to go then.

tmadlener: Yes, that is true. Didn't we agree on "sequential only" for now for all wrapped processors? I am not sure how many would actually properly cope with being in run on multiple threads. How does Gaudi do things in this case?

I think the more important bit here is to get things to run as expected sequentially, otherwise the whole "run processors unchanged" does not really hold, IMHO.

andresailer: Instead of "firstEvent" things like that are done in initialise, holds true for Gaudi, and also most marlin processors.

So far I think we said: thinks can run in parallel, but not in a reproducible way. So this "isFirstEvent" is mostly something we need to document as limitiations.
https://github.com/key4hep/k4MarlinWrapper/blob/master/doc/starterkit/k4MarlinWrapperCLIC/howtoMultithread.md

tmadlener: Yeah, initializing these things before the event loop is how it should be done. [A quick query](https://github.com/search?q=org%3AiLCSoft+isFirstEvent&type=code) at least shows that there are a few existing processors that rely on `isFirstEvent` to work to register some histogram.

I will add this limitation to the documentation.

andresailer: Thanks!

tmadlener:
Updated the documentation to state that `isFirstEvent` does not work in mutli-threaded mode. Should I also add a test that checks that this is working as expected with a simple test processor?

andresailer (approved):
Merge this as well ?

tmadlener:
Yes, unless we want a test for this?

andresailer:
Having a test is always better of course. But can add  one later, to have this in the tag.

PR #95, opened 2022-11-23
Fix leaking LCEvent every loop

tmadlener:
BEGINRELEASENOTES
- Fix memory leak in `LcioEvent`, where the event read from file was leaked every event.

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
+++ b/k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
@@ -22,6 +22,7 @@
  */
 
 #include "k4MarlinWrapper/MarlinProcessorWrapper.h"
+#include "IMPL/LCEventImpl.h"
 
 DECLARE_COMPONENT(MarlinProcessorWrapper)
 
@@ -209,9 +210,9 @@ StatusCode MarlinProcessorWrapper::execute() {
 
   if (sc.isFailure()) {
     // Register empty event
-    the_event = new lcio::LCEventImpl();
     debug() << "Registering empty Event for EDM4hep to LCIO conversion event in TES" << endmsg;
-    auto       pO     = std::make_unique<LCEventWrapper>(the_event, true);
+    auto pO           = std::make_unique<LCEventWrapper>(std::make_unique<lcio::LCEventImpl>());
+    the_event         = static_cast<IMPL::LCEventImpl*>(pO->getEvent());
     StatusCode reg_sc = evtSvc()->registerObject("/Event/LCEvent", pO.release());
     if (reg_sc.isFailure()) {
       error() << "Failed to register empty LCIO Event" << endmsg;
--- a/k4MarlinWrapper/k4MarlinWrapper/LCEventWrapper.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/LCEventWrapper.h
@@ -28,23 +28,18 @@
 
 #include <GaudiKernel/DataObject.h>
 
+#include <memory>
+
 class LCEventWrapper : public DataObject {
 public:
-  // Set delete_event to true when manually creating the LCEvent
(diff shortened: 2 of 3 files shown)


andresailer:
Seems I was wrongly believing that the LCReader keeps ownership of the lcio event.

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -209,14 +209,14 @@ StatusCode MarlinProcessorWrapper::execute() {
...
-    the_event = new lcio::LCEventImpl();
     debug() << "Registering empty Event for EDM4hep to LCIO conversion event in TES" << endmsg;
-    auto       pO     = std::make_unique<LCEventWrapper>(the_event, true);
+    auto       pO     = std::make_unique<LCEventWrapper>(std::make_unique<lcio::LCEventImpl>());
     StatusCode reg_sc = evtSvc()->registerObject("/Event/LCEvent", pO.release());
     if (reg_sc.isFailure()) {
       error() << "Failed to register empty LCIO Event" << endmsg;
       return reg_sc;
     }
+    the_event = static_cast<IMPL::LCEventImpl*>(static_cast<LCEventWrapper*>(pObject)->getEvent());
```
andresailer, line 219: This is causing the test to fail.

In this case here pObject is `nullptr` because we are in the `sc.isFailure()` still

You probably need to store the `LCEventImpl()` pointer already, unless you want to retrieve what was just stored.

tmadlener:
> Seems I was wrongly believing that the LCReader keeps ownership of the lcio event.

I think that changed by going from the `LCReader` returned from the factory to the `MTReader`.

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
tmadlener: The idea was here to just get back the pointer to what I just stored. I figured, since the failure case returns before I get to here, I would always get a valid pointer here, but apparently that is not true?

tmadlener: Ah wait... wrong object...

k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp
```diff
@@ -209,14 +209,16 @@ StatusCode MarlinProcessorWrapper::execute() {
...
     debug() << "Registering empty Event for EDM4hep to LCIO conversion event in TES" << endmsg;
-    auto       pO     = std::make_unique<LCEventWrapper>(the_event, true);
+    auto       pO     = std::make_unique<LCEventWrapper>(std::make_unique<lcio::LCEventImpl>());
     StatusCode reg_sc = evtSvc()->registerObject("/Event/LCEvent", pO.release());
     if (reg_sc.isFailure()) {
       error() << "Failed to register empty LCIO Event" << endmsg;
       return reg_sc;
     }
+    DataObject* eventObject = nullptr;
+    eventSvc()->retrieveObject("/Event/LCEvent", eventObject).ignore();
```
tmadlener, line 220: I suppose ignoring the StatusCode here is something we can do, because just above we make sure that something is actually there?

andresailer: My experience with this is not large enough to know if that can go wrong :)

I think getting `theEvent` out of `pO` before the `registerObject` is more straightforward?
`theEvent = static_cast<IMPL::LCEventImpl*>(pO->getEvent());`

tmadlener: Yeah, that is probably more straight forward, and one less thing to worry about. Will do that

andresailer (approved):

PR #99, opened 2022-11-30
EDM4hep2LCIO: fix call to FillMissingCollections

### LINKED ISSUE #98: Calorimeter hits problem following in CLIC reco recipe
!!!IMPORTANT!!!: To facilitate faster and easier response to your issue please provide in addition to the description of the issue also the following information

 - OS version: centos7
 - key4hep version: nightly tag of 2022/11/28
 - Package version: commit [7e8f0e0](https://github.com/key4hep/k4MarlinWrapper/commit/7e8f0e04e9c2577019332f3ddd0866acb10b26e5)
 - Reproduced by: I have followed the recipe in the k4MarlinWrapperCLIC [web page](https://key4hep.github.io/key4hep-doc/k4marlinwrapper/doc/starterkit/k4MarlinWrapperCLIC/Readme.html)
 - Input: Input files `clicRec_e4h_input.py` and `ttbar_edm4hep.root` in:
`/afs/cern.ch/user/e/ebrondol/public/k4MarlinWrapper/Issue/`
with debug printout `podio_dump_ttbar_edm4hep.log`
 - Output: Output files `my_output.root` and `Output_REC_k4run_edm4hep.slcio` in
`/afs/cern.ch/user/e/ebrondol/public/k4MarlinWrapper/Issue/`
with debug printout `podio_dump_my_output.log` and `anajob_Output_REC_k4run_edm4hep.log`, respectively .
Log file of the running process in `k4run_reco_edm4hep.log`
 - Goal: In the output SLCIO file all calorimeters collections (sim & reco) are empty. In the output edm4hep file the sim calorimeter hits are correctly there, while the reco calo hits are not.
- Additional info: I have tried to test the full reco chain with SLCIO as input and output and it works fine, so the problem is not in the wrapping or content of the single processors.

andresailer:
We have to run some conversions in specific order, or make sure that links are correctly set. This also needs a fix in k4EDM4hep2LcioConv to correctly assign the SimCaloHitContributions

BEGINRELEASENOTES
- EDM4hep2LCIO: Fix so that FillMissingCollections is always called after other collections were converted. Needed also to correctly assign SimCalorimeterHitContributions. Depends on key4hep/k4EDM4hep2LcioConv#7, fixes #98

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -291,11 +291,9 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
     } else {
       debug() << " Collection " << lcioName << " already in place, skipping conversion. " << endmsg;
     }
-
-    if (!m_convertAll) {
-      FillMissingCollections(collection_pairs);
-    }
   }
+
+  FillMissingCollections(collection_pairs);
 
   return StatusCode::SUCCESS;
 }


tmadlener:
We could add a sort of diff based regression test using the outputs of `anajob` or `podio-dump` on the outputs of the clic reco. We would then have to make sure that the reconstruction does not depend on a random seed (or if it does, fix that), and store a "good" output into the tests, and then just run `diff` at the end, to see if things are as expected.

It probably doesn't cover all the possibilities, but it would at least catch the most obvious regressions, like the one that is fixed here. It could also not be entirely robust, so the comparison file would need frequent changes, I can't really judge that. A better test would be to actually compare the LCIO and EDM4hep output directly, but that requires quite a bit more work.

tmadlener (approved):

PR #101, opened 2022-12-07
Add diff based tests to validate clic reconstruction outputs

tmadlener:
BEGINRELEASENOTES
- Add `diff` based tests that compare the output files of the clic reconstruction test, with expected outputs via `anajob` and `podio-dump`. This should catch very obvious regressions like #98, already in CI.
- Remove the restriction of running the clic reconstruction test only in nightly stacks.

ENDRELEASENOTES

This is probably not the most robust setup, but it should at least stop very obvious regressions, and updating the expected outputs from time-to-time also gives us some sort of bread crumbs to see what things changed when.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -58,12 +58,7 @@ if (BASH_PROGRAM)
       PASS_REGULAR_EXPRESSION "Input and output have same number of events")
 
   # Test clicReconstruction with EDM4hep input and output
-  # Disabled for current version in k4lcioreader
-  get_target_property(k4lcioreader_PATH  k4LCIOReader::k4LCIOReader LOCATION)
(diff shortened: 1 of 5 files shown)


tmadlener:
Maybe we want to hold off on the `podio-dump` bit until AIDASoft/podio#352 is fixed? That would save us one update of the corresponding `.expected` file.

tmadlener:
I think the `sw.hsf.org` based workflow still suffers from #98 (judging from the outputs of `anajob` in the diff)

andresailer:
> I think the `sw.hsf.org` based workflow still suffers from #98 (judging from the outputs of `anajob` in the diff)

The test works! :)

tmadlener:
Shall we merge this and make tag in order to pick this up in the releases? Then we should have all workflows working again (hopefully).

andresailer (approved):

PR #103, opened 2022-12-21
put LCIO input files into global::parameters

gaede:
BEGINRELEASENOTES
- LcioEventAlgo: put LCIO input files into global::parameters
   - needed e.g. by PatchCollection processor
   - done in LcioEventAlgo::init()
   - this fix is needed to be able to convert LCIO files to EDM4hep when not all collections are present in all events (default in ILD reconstruction)
- ensure marlin::Global:parameters is created exactly once

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/util/k4MarlinWrapperUtil.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/util/k4MarlinWrapperUtil.h
@@ -24,11 +24,20 @@ namespace k4MW::util {
     return split(subject, re);
   }
 
-  /// singleton helper to intialise global Marlin parameters exactly once
-  const static marlin::StringParameters* marlin_global_parameters = ([]() {
-    marlin::Global::parameters = new marlin::StringParameters;
+  /// singleton helper to initialize global Marlin parameters exactly once. This
+  marlin::StringParameters* marlinGlobalParameters() {
+    static marlin::StringParameters p{};
+    const static auto               initMarlinGlobal = []() {
+      marlin::Global::parameters = &p;
+      return true;  // need a non-void return type
+    }();
+
     return marlin::Global::parameters;
-  })();
+  }
+
+  // Make sure that the initialization is done before any possible accesses to
+  // marlin::Global::parameters
+  const static auto* marlin_global_parameters = marlinGlobalParameters();
 
 }  // namespace k4MW::util
 
--- a/k4MarlinWrapper/k4MarlinWrapper/util/k4MarlinWrapperUtil.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/util/k4MarlinWrapperUtil.h
@@ -6,6 +6,7 @@
 #include <string>
 #include <vector>
 
+#include "marlin/Global.h"
 #include "marlin/StringParameters.h"
 
 namespace k4MW::util {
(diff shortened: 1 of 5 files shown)


tmadlener (commented):
Very nice, I would have thought this to be more involved. From my point of view this looks good, and even fixes a small leak(?). The CI is not yet happy due to some formatting issues.

**Very technically** the two assignments to `Marlin::Global::parameters` could race, and it is currently still the responsibility of the "user" to assign the `StringParameters` to the `marlin::Global::parameters`. I think this assignment could be moved into the `globalParameters` function (and made to happen only once). That would have the advantage of being thread safe (at least the assignment of the `parameters`), and the using side could essentially simply do, e.g.
```cpp
k4MW::util::globalParameters()->add("LCIOInputFiles", m_fileNames);
```
not having to worry about assignment and saving a whole line for each of the two calls ;)

The implementation would then look something like this:
```cpp
marlin::StringParameters* globalParameters() {
  static marlin::StringParameters p{};
  const static auto initMarlinGlobal = [](){
    marlin::Global::parameters = &p;
    return (marlin::Global::parameters != nullptr); // or simply true
  }();
  return marlin::Global::parameters;
}
```

NB: one could also could ditch the `globalParameters` function entirely and just keep the `static StringParameters` and the immediately invoked lambda, and use `marlin::Global::parameters` directly in using code.

gaede:
Hi Thomas, nice trick w/ the IIFE.  How do I get the required formatting done ?

k4MarlinWrapper/k4MarlinWrapper/util/k4MarlinWrapperUtil.h
```diff
@@ -20,6 +23,13 @@ namespace k4MW::util {
     std::regex re{"\\s+"};
     return split(subject, re);
   }
+
+  /// singleton helper to intialise global Marlin parameters exactly once
+  const static marlin::StringParameters* marlin_global_parameters = ([]() {
+    marlin::Global::parameters = new marlin::StringParameters;
+    return marlin::Global::parameters;
+  })();
```
tmadlener, line 31: This is unfortunately apparently not enough to guarantee that the global parameters are initialized only (and exactly) once. This is also causing the test to fail, as this initialization now happens twice, therefore:
- the global parameters no longer have a `RandomSeed` stored
- `StringParameters::getIntVal` returns `0` if a key is not found

In this way the random seed is reset to a different value, causing the test to fail.

tmadlener: OK. The reason this is not working as expected is because a global static variable signifies a variable that is only visible inside one translation unit. Given that this header is included in two translation units, we have a double initialization. Function scope statics on the other hand will always refer to the same variable, which is why the initialization as implemented now works.

andresailer: You fixed this issue, and things are OK now as is, right?

tmadlener: Yes, that is correct. The `marlin::Global::parameters` are now correctly initialized (thread safe even), and as long as the `k4MarlinWraperUtil.h` header is included in a translation unit, they can be used directly anywhere (that on the other hand is only thread safe for reading)

andresailer (approved):

PR #106, opened 2023-01-12
Make it possible to run clic reco tests concurrently

tmadlener:
BEGINRELEASENOTES
- Make it possible to run clic reco tests concurrently, by ensuring that cloning the CLICPerformance repository is not done concurrently by several test processes.
- Only get the last commit of CLICPerformance to speed up the cloning.

ENDRELEASENOTES


andresailer (approved):
I was thinking if FetchContent would be better to clone the clicperformance repo. But this let's one still run the scripts standalone as well.

tmadlener:
Ah, I hadn't even thought about FetchContent tbh. I figured the script would be the easiest solution to only fetch when necessary, i.e. when actually running the tests (instead of at configuration time with FetchContent).

PR #107, opened 2023-02-10
Fix link in documentation

tmadlener:
BEGINRELEASENOTES
- Fix link to edm converters documentation

ENDRELEASENOTES

Reported broken by CI of key4hep-doc: https://github.com/key4hep/key4hep-doc/actions/runs/4133418708/jobs/7164050580#step:6:118


tmadlener:
doctest is failing because the key4hep nightlies come with ddsim that already produces `podio::Frame` output

PR #111, opened 2023-03-28
Transfer EventHeader information to LCIO event

### LINKED ISSUE #110: Issues when running CLIC reconstruction on EDM4Hep input
_TL;DR:
event and run number not set/transferred during edm4hep to lcio conversion?
/TL;DR_

We received a bug report via email from Emmanuel Perez (original message below) about inconsistencies in the CLIC reconstruction between using Marlin+LCIO, Gaudi+LCIO and Gaudi+EDM4Hep. 
He noticed a shift in the [track pulls](https://github.com/key4hep/k4MarlinWrapper/files/11059607/summary_track_pulls.pdf) when using the edm4hep input, mainly in tanLambda, z0 and d0. I was able to reproduce this following his description. 

Digging into this we noticed that there must be an issue already during digitisation as the hit position smearing is non-gaussian.

with edm4hep input:

with slcio input:

which is weird as [this part of the code](https://github.com/iLCSoft/MarlinTrkProcessors/blob/master/source/Digitisers/src/DDPlanarDigiProcessor.cc) looks to be independent from the input:
```
void DDPlanarDigiProcessor::init() { 
  //...
  _rng = gsl_rng_alloc(gsl_rng_ranlxs2);
  Global::EVENTSEEDER->registerProcessor(this);
  //...
}

void DDPlanarDigiProcessor::processEvent( LCEvent * evt ) { 

  gsl_rng_set( _rng, Global::EVENTSEEDER->getSeed(this) ) ;   
  //...

    //later in a loop over the hits:
    double uSmear  = gsl_ran_gaussian( _rng, resU ) ;
    _h[hu]->Fill(  uSmear / resU ) ; 
}
```
However, there is a dependency on the event during the refreshment of the seeds happening for every event and as far as I can tell the event and run numbers are not set in the edm4hep to lcio conversion?

```
void ProcessorEventSeeder::refreshSeeds( LCEvent * evt ) {
  //...
  // get hashed seed using jenkins_hash
  unsigned int seed = 0 ; // initial state
  unsigned int eventNumber = evt->getEventNumber() ;
  unsigned int runNumber = evt->getRunNumber() ;

  unsigned char * c = (unsigned char *) &eventNumber ;
  seed = jenkins_hash( c, sizeof eventNumber, seed) ;

  c = (unsigned char *) &runNumber ;
  seed = jenkins_hash( c, sizeof runNumber, seed) ;

  c = (unsigned char *) &_global_seed ;
  seed = jenkins_hash( c, sizeof _global_seed, seed) ;
  //...
  srand( seed );
  //...
}
```
I will check that next...

> Dear Thomas and André,
> 
> Thanks for setting up the nice doc for the k4MarlinWrappers :
> https://github.com/key4hep/k4MarlinWrapper/tree/master/doc/starterkit/k4MarlinWrapperCLIC
> 
> I see issues with the "edm4hep workflow" - produce an edm4hep file with ddsim, and reconstruct it with Gaudi via the wrappers.
> 
> I have simulated 500  Z -> mumu events (at 91 GeV), with ddsim, with the CLIC_o3_v14.xml model. I ran ddsim twice, first to produce an slcio file, and second, to produce an edm4hep file.
> Then, I have run the reconstruction, following the instructions. I have reconstructed the slcio file both with Marlin and with Gaudi.
> The reconstruction runs the "TrackChecker" module (*), which produces histograms of the track quantities, in particular the pulls of the track parameters. The full set of commands is given below (**).
> 
> The attached pdf shows these pulls of the track parameters, for the three workflows. The corresponding root files are on lxplus in ~eperez/public/for_reco
> 
> The pulls are in general a bit large, even in the "historical" workflow (slcio + Marlin), especially for the omega parameter, but maybe that's OK ?
> What worries me are the *shifts* that are observed in the plots coming from the edm4hep workflow, which are most pronounced in tanLambda, d0 and in particular z0.
> Since these shifts are not seen when I reconstruct the slcio events with gaudi, I assume that the problem lies in the edm4hep to lcio conversion.
> 
> Do you have an idea fo what is going wrong ?
> 
> Thanks a lot and cheers,
> E.
> 
> 
> (*) for the edm4hep workflow: I have had to update the collection name from "MCParticle" to "MCParticles" in the definition of MyTrackChecker (and of MyClicEfficiencyCalculator), in the example that you indicate in the k4MarlinWrapper repository, test/gaudi_opts/clicRec_e4h_input.py
> 
> (**)
> source /cvmfs/sw.hsf.org/key4hep/setup.sh
> 
> edm4hep workflow:
> ddsim --compactFile $LCGEO/CLIC/compact/CLIC_o3_v14/CLIC_o3_v14.xml --outputFile Zmumu_edm4hep.root --steeringFile clic_steer.py --inputFiles /eos/experiment/fcc/ee/generation/hepmc/p8_ee_Zmumu_ecm91/events_noVtxSmear.hepmc --numberOfEvents 500
> k4run clicRec_e4h_input.py
> 
> slcio + marlin :
> same ddsim command  but with --outputFile Zmumu.slcio
> Marlin clicReconstruction.xml --InitDD4hep.DD4hepXMLFile=$LCGEO/CLIC/compact/CLIC_o3_v14/CLIC_o3_v14.xml --global.LCIOInputFiles=Zmumu.slcio --global.MaxRecordNumber=500
> 
> slcio + gaudi :
> k4run clicReconstruction.py
> 
> The various .py and .xml files I used are on lxlpus, in  ~eperez/public/for_reco

Zehvogel:
BEGINRELEASENOTES
- edm4hep2lcio conversion: Transfer run and event number from edm4hep EventHeader to LCIO event. Fixes #110 

ENDRELEASENOTES

There are still some TODOs in here where I was not sure what the best option to warn the user would be.

Depends on https://github.com/key4hep/k4EDM4hep2LcioConv/pull/13

### DIFF
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -213,15 +213,14 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
 }
 
 // Transfer info from EDM4hep EventHeader to LCIO event
-void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event)
-{
+void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event) {
   // ReconstructedParticles handle
   DataHandle<edm4hep::EventHeaderCollection> header_handle{e4h_coll_name, Gaudi::DataHandle::Reader, this};
-  const auto header_coll = header_handle.get();
+  const auto                                 header_coll = header_handle.get();
 
-  const auto event_n = header_coll->eventNumber();
-  const auto run_n = header_coll->runNumber();
-  const auto timestamp = header_coll->timeStamp();
+  const auto event_n      = header_coll->eventNumber();
+  const auto run_n        = header_coll->runNumber();
+  const auto timestamp    = header_coll->timeStamp();
   const auto event_weight = header_coll->weight();
 
   // the collection returns vectors but they should be of length 1
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -217,6 +217,10 @@ void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio
   DataHandle<edm4hep::EventHeaderCollection> header_handle{e4h_coll_name, Gaudi::DataHandle::Reader, this};
   const auto                                 header_coll = header_handle.get();
 
+  if (header_coll->size() != 1) {
+    error() << "Header collection contains " << header_coll->size() << " headers, expected 1." << endmsg;
+    return;
+  }
   convEventHeader(header_coll, lcio_event);
 }
 
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -306,7 +306,6 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
     }
   }
 
-  // TODO: warn here if EventHeader not there
   debug() << "Event: " << lcio_event->getEventNumber() << " Run: " << lcio_event->getRunNumber() << endmsg;
 
   FillMissingCollections(collection_pairs);
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -214,17 +214,16 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
 
 // Transfer info from EDM4hep EventHeader to LCIO event
 void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event) {
-  // ReconstructedParticles handle
   DataHandle<edm4hep::EventHeaderCollection> header_handle{e4h_coll_name, Gaudi::DataHandle::Reader, this};
   const auto                                 header_coll = header_handle.get();
 
-  const auto event_n      = header_coll->eventNumber();
-  const auto run_n        = header_coll->runNumber();
-  const auto timestamp    = header_coll->timeStamp();
-  const auto event_weight = header_coll->weight();
+  const auto& event_n      = header_coll->eventNumber();
+  const auto& run_n        = header_coll->runNumber();
+  const auto& timestamp    = header_coll->timeStamp();
+  const auto& event_weight = header_coll->weight();
 
   // the collection returns vectors but they should be of length 1
-  if (event_n.size() != 1 || run_n.size() != 1) {
+  if (event_n.size() != 1 || run_n.size() != 1 || timestamp.size() != 1 || event_weight.size() != 1) {
     // TODO: fail harder?
     error() << "Malformed EventHeader, multiple entries for event number, run number, timestamp or weight!" << endmsg;
     return;
@@ -319,7 +318,7 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
   }
 
   // TODO: warn here if EventHeader not there
-  info() << "Event: " << lcio_event->getEventNumber() << " Run: " << lcio_event->getRunNumber() << endmsg;
+  debug() << "Event: " << lcio_event->getEventNumber() << " Run: " << lcio_event->getRunNumber() << endmsg;
 
   FillMissingCollections(collection_pairs);
 
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -212,6 +212,31 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
   lcio_event->addCollection(conv_recops, lcio_coll_name);
 }
 
+// Transfer info from EDM4hep EventHeader to LCIO event
+void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event)
+{
+  // ReconstructedParticles handle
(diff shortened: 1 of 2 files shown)


k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -212,6 +212,30 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
...
+  // the collection returns vectors but they should be of length 1
+  if (event_n.size() != 1 || run_n.size() != 1) {
+    // TODO: fail harder?
+    error() << "Malformed EventHeader, multiple entries for event number, run number, timestamp or weight!" << endmsg;
+    return;
+  }
+
+  lcio_event->setEventNumber(event_n[0]);
+  lcio_event->setRunNumber(run_n[0]);
+  lcio_event->setTimeStamp(timestamp[0]);
```
andresailer, line 235: At the moment there are only checks for eventnumber and runnumber, so timestamp and event_weight might be empty.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -293,6 +318,9 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
     }
   }
 
+  // TODO: warn here if EventHeader not there
+  info() << "Event: " << lcio_event->getEventNumber() << " Run: " << lcio_event->getRunNumber() << endmsg;
```
andresailer, line 322:
```suggestion
  info() << "Event: " << lcio_event->getEventNumber() << " Run: " << lcio_event->getRunNumber() << endmsg;
```

better some other verbosity level (debug? verbose?)

andresailer (commented):
nicely spotted the problem!

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -212,6 +212,30 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
   lcio_event->addCollection(conv_recops, lcio_coll_name);
 }
 
+// Transfer info from EDM4hep EventHeader to LCIO event
+void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event) {
+  // ReconstructedParticles handle
```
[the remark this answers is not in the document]
Zehvogel: Yes, I will remove that comment.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
Zehvogel: Thanks for spotting this

tmadlener:
Should we consider moving the actual "conversion" to k4EDM4hep2LCIOConv and simply put the necessary Gaudi wrapper code in here?

andresailer:
> Should we consider moving the actual "conversion" to k4EDM4hep2LCIOConv and simply put the necessary Gaudi wrapper code in here?

Yes, you are correct, we should put the "conversion" in k4EDM4hep2LCIOConv and call it from here.

Zehvogel:
could you specificy which part of the "conversion" I should move?  Should the function in k4EDM4hep2LCIOConv return the 4 values or also set them in the LCIO event? The latter would be contrary to how the other functions are while the former seems overly complicated?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -212,6 +212,29 @@ void EDM4hep2LcioTool::convertReconstructedParticles(
...
 
+// Transfer info from EDM4hep EventHeader to LCIO event
+void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio::LCEventImpl* lcio_event) {
+  DataHandle<edm4hep::EventHeaderCollection> header_handle{e4h_coll_name, Gaudi::DataHandle::Reader, this};
+  const auto                                 header_coll = header_handle.get();
+
+  const auto& event_n      = header_coll->eventNumber();
+  const auto& run_n        = header_coll->runNumber();
+  const auto& timestamp    = header_coll->timeStamp();
+  const auto& event_weight = header_coll->weight();
```
tmadlener, line 223: You are calling the "vectorized access functionality here". It would be much easier to simply to make sure the collection has only one element and then retrieve the first element from the collection and use the `get` prefixed functions to handle only one event header, sth like:
```cpp
lcio_event->setEventNumber(evtHeader.getEventNumber());
// ...
```

tmadlener (commented):
The "conversion" in this case is effectively the 4 lines of setting information in the LCEvent. The interface in k4EDM4hep2LcioConv should probably looks something like:
```cpp
void setEventHeaderInfo(lcio::LCEventImpl* evt, const edm4hep::EventHeaderCollection* const evtHeaderColl);
```
so that the k4MarlinWrapper side of this would effectively boil down to getting the handle and passing it on.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
Zehvogel: thanks, I was not aware of that possibility, much easier indeed :)

Zehvogel:
I made the suggested changes and also opened https://github.com/key4hep/k4EDM4hep2LcioConv/pull/13

tmadlener:
Unrelated clang-format issues should be fixed with #112

Zehvogel:
I moved the size check + error message to this side now and got rid of the unrelated format changes

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -293,6 +306,9 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
     }
   }
 
+  // TODO: warn here if EventHeader not there
```
tmadlener, line 309: Is TODO item still relevant?

Zehvogel: No, I removed it now

tmadlener (approved):
Thanks. Looks good to me. I would just like to have at least one successful build in CI before merging this.

Zehvogel:
Can we get this merged? It builds successfully in the nightlies and I think the failing test is also failing on master and caused by https://github.com/key4hep/k4MarlinWrapper/issues/113.

PR #112, opened 2023-04-18
[clang-format] Fix formatting

tmadlener:
BEGINRELEASENOTES
- Fix formatting

ENDRELEASENOTES

Fixing formatting in a separate PR here to get the `clang-format-check` workflow working properly again and to declutter other PRs

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -65,8 +65,9 @@ private:
                                  const std::string& e4h_coll_name, const std::string& lcio_coll_name,
                                  lcio::LCEventImpl* lcio_event);
 
-  void convertTPCHits(vec_pair<lcio::TPCHitImpl*, edm4hep::RawTimeSeries>& tpc_hits_vec, const std::string& e4h_coll_name,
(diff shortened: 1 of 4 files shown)


andresailer:
The error in the edm_coverters test is caused by the LCIO Tracks being converted before the LCIO TrackerHit collection, and then there are no TrackerHits in the the EDM4hep Track Collection

andresailer:
This https://github.com/key4hep/k4LCIOReader/pull/28 is causing the problem in the edm_converters

tmadlener:
Hasn't this been working with that changes in place already in the past?

andresailer:
At least the test probably hasn't worked. If I revert that commit and run the test, the test passes.

andresailer:
What I don't understand is that even with
https://github.com/key4hep/k4MarlinWrapper/blob/b2b6816091c95046e2b03b3a5105ede6da530a8c/test/gaudi_opts/edm_converters.py#L30
the test passes with the changes in k4LCIOReader I propose in https://github.com/key4hep/k4LCIOReader/pull/30

tmadlener:
Since the long-term solution would be to replace the `k4LCIOReader` here with key4hep/k4EDM4hep2LcioConv#11 in any case, should we try that already now?

andresailer (approved):
Since this is just changing format, and the test error is hopefully solved, let's merge

PR #120, opened 2023-07-06
Replace the Legacy DataSvc and use the new Frame based one

tmadlener:
BEGINRELEASENOTES
- Switch the converters to use the new Frame based `DataSvc` components introduced in [key4hep/k4FWCore#100](https://github.com/key4hep/k4FWCore/pull/100)
- Use the `MetaDataHandle` for the `CellIDEncoding` string passing in the converters
- Update the test inputs and expected outputs in order to be able to work with these changes 
- Update the documentation to also include the "workaround" for [key4hep/k4FWCore#125](https://github.com/key4hep/k4FWCore/issues/125)

ENDRELEASENOTES

Needs
- [x] AIDASoft/podio#439
- [x] key4hep/k4FWCore#119
- [x] key4hep/k4FWCore#124
- [x] New input file for the `clicRec_edm4hep_input` test (needs to go to the cernbox from where it is downloaded)

Includes changes from
- [x] #114 
- [x] #118 

TODO
- [x] Release Notes (waiting to see how this goes once all the bits and pieces are in place)
- [x] Remove `k4LCIOReader` dependency
- [x] Update tests (provide new input file)


tmadlener:
I opened key4hep/k4FWCore#125 to keep track of the non-zero (4) exit code of `k4run` and switched to simply ignoring the return value in the edm4hep converter test for now. It shouldn't really matter since there are checks afterwards that will catch problems, just debugging the test will become a bit more cumbersome.

doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
```diff
@@ -102,13 +102,17 @@ ApplicationMgr( TopAlg = algList,
               )
 ```
 
-With all these pieces put together (as in `examples/event_display.py`) it is no possible to run the event display. In order to run the event display we first have to start the `glced` server program to which the wrapped `CEDViewer` processor will then connect. Starting the server and running the wrapped processor can be done via
+With all these pieces put together (as in `examples/event_display.py`) it is now possible to run the event display. In order to run the event display we first have to start the `glced` server program to which the wrapped `CEDViewer` processor will then connect. Starting the server and running the wrapped processor can be done via
 ```bash
 glced &
 
-k4run $K4MARLINWRAPPER/examples/event_display.py --EventDataSvc.input=gamma_10GeV_edm4hep.root
+k4run $K4MARLINWRAPPER/examples/event_display.py --EventDataSvc.input=gamma_10GeV_edm4hep.root || true
```
jmcarcell, line 109: I would prefer this to be opt-in: we give the command without `|| true` and then an explanation on why you may want to use it if running in a test or something like that, but that's kind of expert knowledge that is unlikely to be useful for most people

doc/starterkit/k4MarlinWrapperCLIC/Readme.md
```diff
@@ -124,14 +124,18 @@ included with k4MarlinWrapper. Note that:
...
+`k4run` exit with a non-zero exit code even though nothing went wrong.
 
 ```bash
 cd CLICPerformance/clicConfig
 
 cp ../../../test/gaudi_opts/clicRec_e4h_input.py .
 
 k4run clicRec_e4h_input.py --EventDataSvc.input ttbar_edm4hep.root
+result=$?
+[ $result = "0" ] || [ $result = "4" ] && true
```
jmcarcell, line 139: Same comment as above, not needed for most people so it can be separated from the code that is going to be copied and pasted most of the time

jmcarcell (approved):

doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
tmadlener: The problem is that these bits of the documentation are run in the `doctest` workflow and without any handling of the non-zero exit code from `k4run` they will fail.

PR #123, opened 2023-07-19
Don't use run to run k4run

jmcarcell:
BEGINRELEASENOTES
- Remove run (autogenerated script to set up the environment) from some tests that use it together with k4run; we should set up the environment either in `CMakeLists.txt` or externally. 

ENDRELEASENOTES

Either we set the environment externally or in CMakeLists.txt but don't use run and force some relative paths that may or may not be how it is built. I tested this and both test work fine

### DIFF
--- a/test/scripts/clicRec_lcio_mt.sh
+++ b/test/scripts/clicRec_lcio_mt.sh
@@ -17,7 +17,7 @@ if [ ! -f clicReconstruction_mt.py ]; then
   ln -s $TEST_DIR/gaudi_opts/clicReconstruction_mt.py clicReconstruction_mt.py
 fi
 
-../../../run k4run clicReconstruction_mt.py
+k4run clicReconstruction_mt.py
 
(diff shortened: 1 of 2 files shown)


andresailer:
Can you please elaborate more on the release notes?

jmcarcell:
Done. CI is failing but the two tests that this is changing succeeded

tmadlener:
who or what generates the `run` script? If it is no longer necessary can it be removed?

jmcarcell:
That comes from Gaudi: https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/cmake/GaudiToolbox.cmake#L1546
Probably people rely on this so it's not easy to have it removed. It doesn't hurt either, it's just there if we don't use it.
This is the only issue I found about this: https://gitlab.cern.ch/gaudi/Gaudi/-/issues/183

jmcarcell:
Ahah now one the tests that succeeded in the checks in this PR and also after https://github.com/key4hep/k4MarlinWrapper/pull/114 fails, of course

PR #133, opened 2023-07-31
Only link to libraries that are actually needed

jmcarcell:
BEGINRELEASENOTES
- Hardcode the LCIO and Marlin libraries to only link to the libraries that are actually needed

ENDRELEASENOTES
For example Marlin builds a library called MarlinXML that libraries here don't need to link to

### DIFF
--- a/k4MarlinWrapper/CMakeLists.txt
+++ b/k4MarlinWrapper/CMakeLists.txt
@@ -27,7 +27,7 @@ gaudi_add_module(k4MarlinWrapperPlugins
     src/components/LcioEventOutput.cpp
   LINK
     Gaudi::GaudiAlgLib
-    ${Marlin_LIBRARIES}
+    Marlin
 )
 
 target_include_directories(k4MarlinWrapperPlugins PUBLIC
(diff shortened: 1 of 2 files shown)


k4MarlinWrapper/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ gaudi_add_module(k4MarlinWrapperPlugins
     src/components/LcioEventOutput.cpp
   LINK
     Gaudi::GaudiAlgLib
-    ${Marlin_LIBRARIES}
+    Marlin
```
andresailer, line 30: This just makes cmake add `-lMarlin`, so you would have to add `-L /path/to/marlin/lib` as well.

Or you make Marlin create targets.

Or you add `-Wl,--as-needed`
But make sure that doesn't drop libraries that we need for the EDM at runtime, but not at linktime?

jmcarcell: Yeah forget about this PR for now, I should just close it and have to find how to solve my issues in a different way I think

PR #135, opened 2023-08-30
EDM4hep Test: create new file, adjust expectations

andresailer:
BEGINRELEASENOTES
- CI: clicrec_edm4hep test: created new edm4hep SIM file compatible with current podio master

ENDRELEASENOTES


jmcarcell:
I guess this doesn't go into the spack dependencies since it's just the CI and it's not so easy or not worth it to know which version was causing the previous problems
``` python
    depends_on("edm4hep@0.4.1:", when="@0.4.1:")
    depends_on("edm4hep@0.8:", when="@0.6:")
```

andresailer:
The previous file was done with podio 0.16.06+x, but before the master was calling itself 0.16.99.
But we need podio 0.16.07 in FWCore at the moment anyway, as soon as the new podio version comes out.

PR #137, opened 2023-09-04
Fix a comment and check for the status code of a test

jmcarcell:
BEGINRELEASENOTES
- Docs: Correct the comment referring to the input file that has to be replaced
- Test: no longer ignore the return value of a test. It used to return 4, but that was fixed in  key4hep/k4fwcore#132
- Tests: Replace environment variable LCGEO with K4GEO following the usage in other key4hep repositories   

ENDRELEASENOTES

The line that was previously going to be substituted doesn't exist. There is an extra place where LCGEO appears: https://github.com/key4hep/k4MarlinWrapper/blob/main/doc/starterkit/k4MarlinWrapperCLIC/Readme.md?plain=1#L118
but that also requires changes in `CLICPerformance/clicConfig`


andresailer:
Please put in more context for "Change LCGEO to K4GEO"

andresailer (approved):

PR #140, opened 2023-09-13
Remove workarounds for catching non-zero exit codes

tmadlener:
BEGINRELEASENOTES
- Remove the workarounds for non-zero exit codes again. They were introduced in #120 to work around key4hep/k4FWCore#125 which has been fixed with key4hep/k4FWCore#132

ENDRELEASENOTES


andresailer (approved):
Thanks!

PR #142, opened 2023-09-13
Introduce a global object map that is shared by all converter instances

### LINKED ISSUE #113: Broken associations when using edm4hep input
For details see https://github.com/key4hep/k4LCIOReader/issues/31

Maybe this is not worth fixing and we should directly switch to https://github.com/key4hep/k4EDM4hep2LcioConv/pull/11? What is the status on that?

tmadlener:
BEGINRELEASENOTES
- Introduce a *global object map* that is shared by all converter instances (in either direction). This is necessary to properly set all relations even if the related objects are not converted in the same converter instance.
  - Fixes [key4hep/k4LCIOReader#31](https://github.com/key4hep/k4LCIOReader/issues/31)
  - Fixes #113 
- Add a `global_converter_maps` test case (and a description) that checks that this works as expected.

ENDRELEASENOTES

This should solve #113 in the end.

- [ ] Needs key4hep/k4EDM4hep2LcioConv#21
- [x] ~Includes #143 (can be removed if that is not merged)~
- [ ] Includes #144 
- [x] Resetting the global maps for each event.
- [x] Release notes
- [x] Tests (added "minimal" test case + description)


tmadlener:
Things should be running now, but some of the tests are currently failing. I am not yet sure whether the tests are checking for now no longer correct behavior, or whether these are still actual failures.

### `edm_converters`

The "test case" that is failing is the roundtrip conversion from EDM4hep -> LCIO -> EDM4hep of `SimTrackerHits`. The failing part is the comparison of the relations to the MC particles. The test tries to ensure that the MCParticles of the roundtrip converted SimTrackerHits still point to the original MC particles. However, they actually point to the converted MC particles. I am not sure what would be the expected behavior in this case. I am also not sure whether it is easily possible to actively choose either behavior...

tmadlener:
This now seems to solve (almost) all issues that were previously present in the `clicRec_edm4hep_input` test case. All log outputs of the form

```
[...]
Cannot find corresponding EDM4hep object for an LCIO object in a subset collection of type edm4hep::MCParticle
[...]
Cannot find corresponding EDM4hep Tracks for a LCIO Tracks, while trying to resolve the ReconstructedParticle Relations
[...]
Couldn't find CaloHit to add to Relations for Clusters in edm
[...]
Cannot find corresponding EDM4hep object for an LCIO object in a subset collection of type edm4hep::ReconstructedParticle
[...]
Cannot find corresponding EDM4hep object for an LCIO object in a subset collection of type edm4hep::Cluster
[...]
```

are now gone with the exception of a few vertices that still cannot find their corresponding reco particle. Here we might just be missing the conversion of a collection(?). I will still try to add a "minimal" (as minimal as possible) smaller test case that just tests the global map instead of everything in the reconstructoin.

jmcarcell:
Unintended merge that is already reverted

PR #143, opened 2023-09-15
Add some more outputs to the converter tests

tmadlener:
BEGINRELEASENOTES
- Add log outputs to the converter tests to make the (start of the) debugging process a bit quicker.

ENDRELEASENOTES

I have been adding these in quite a few of the recent developments. They are far from complete, but at least for the cases I encountered, they now give you a much better starting point to for debugging than previously.


tmadlener:
No longer necessary with #144

PR #145, opened 2023-09-28
Remove the TrackingCellIDEncodingSvc

jmcarcell:
BEGINRELEASENOTES
- Remove the TrackingCellIDEncodingSvc from k4MarlinWrapper to avoid having a dependency on Marlin to use this service. Move it to k4Reco

ENDRELEASENOTES

Moved to k4Reco in PR:


andresailer:
But then you depend on LCIO in k4Reco? Which would be nice to avoid?

PR #155, opened 2023-10-30
Changes for included files

pandreetto:
This is a simple patch for including part-files in the main steer file.
Unfortunately part-files are not xml compliant and should be included as raw ones.
An example of fragmented configuration is available at https://github.com/MuonColliderSoft/lcgeo/tree/master/MuColl/MuColl_v1

### DIFF
--- a/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
+++ b/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py
@@ -268,14 +268,17 @@ def escapeIllegalChars(file_str):
   return file_str.replace("&&", "&amp;&amp;")
 
 
-def includeFiles(tree):
+def includeFiles(tree, infile):
   for p_item in tree.findall('.//include/..'):
     for if_item in p_item.findall('./include'):
       try:
-        with open(if_item.attrib['ref'], "r+") as iFile:
+        #TODO resolve path for part-files
+        nested_file = if_item.attrib['ref']
+        with open(nested_file, "r+") as iFile:
           # workaround: part files are not xml compliant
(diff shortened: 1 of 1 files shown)


Zehvogel:
Did you mean to add another link? You are not pointing to a steering file but to a geometry description.

pandreetto:
Sorry, that's the wrong reference.
This is the example of configuration for a reconstruction: https://github.com/MuonColliderSoft/MuonCutil/tree/master/SoftCheck
The main file is the [reco_steer.xml](https://github.com/MuonColliderSoft/MuonCutil/blob/master/SoftCheck/reco_steer.xml)

pandreetto:
I understand.
Feel free to reject the pull request.

jmcarcell:
With these includes the only thing Marlin does is paste the contents? If so I think it would be fine to have something like that here if people need it, but it may not be as trivial; for example what if one converts the steering file from a different directory from the one where the main .xml is, are the paths in the includes resolved correctly? 
By the way, what if there is an include inside an included file? That wouldn't work in the current implementation.

tmadlener:
As discussed during [todays Key4hep meeting](https://indico.cern.ch/event/1342144/) we will merge #148  first to make a tag and a new Key4hep release and then pick this up later once we have sorted out some more of the details of the Marlin XML parsing.

One open question that came up today as well was whether Marlin supports `include` statements that contain `constant`s, as the converter script would then also have to resolve those before looking at paths. I will try to collect a little test-suite of real world cases that we would need to support.

jmcarcell:
Is this still needed @pandreetto?

pandreetto:
This request is not necessary anymore

PR #156, opened 2023-11-01
Use the new edm4hep::CellIDEncoding

jmcarcell:
BEGINRELEASENOTES

- Use the new `edm4hep::CellIDEncoding`. Needs https://github.com/key4hep/EDM4hep/pull/234

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ include(GNUInstallDirs)
 find_package(Gaudi REQUIRED)
 find_package(LCIO REQUIRED)
 find_package(Marlin REQUIRED)
-find_package(EDM4HEP REQUIRED)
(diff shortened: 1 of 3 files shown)


tmadlener:
Using this as a proxy for all of the similar PRs. Do we want these in the release already? (Would require re-tagging a few packages). In any case we will have to wait for the nightlies to pick up the latest EDM4hep.

andresailer:
> Using this as a proxy for all of the similar PRs. Do we want these in the release already? (Would require re-tagging a few packages). In any case we will have to wait for the nightlies to pick up the latest EDM4hep.

We should have the fix for the incompatible variant. For the other cases I wouldn't bother tagging just for that. But here in k4MarlinWrapper we anyway still have another bug fix merged so we have to make a new tag.

tmadlener (approved):

PR #159, opened 2023-11-04
Add the include directory at the top of EDM4hep

jmcarcell:
BEGINRELEASENOTES
- Add the include directory at the top of EDM4hep for builds when EDM4hep and k4MarlinWrapper are built together and the `include` directory has not been installed yet and can't be used. For builds where EDM4hep is installed before `k4MarlinWrapper` the behavior is unchanged.

ENDRELEASENOTES


andresailer:
I think the proper solution for this is to add the proper `<BUILD_INTERFACE:>` and `<INSTALL_INTERFACE:>` directories to the EDM4hep targets.

Issue #153, opened 2023-10-23
Always convert EventHeader collection


Zehvogel:
otherwise one can forget about it and then event numbers etc. are missing leading to the same symptoms as https://github.com/key4hep/k4MarlinWrapper/issues/110

tmadlener:
Can you be a bit more specific on what the issue is, or how to reproduce it? In principle the EventHeader should be converted unless this information is already there. However, it seems that this is not working as intended, or that something is missing?

Zehvogel:
Copying how to read and convert edm4hep files from the below test leads to missing event numbers as the EventHeader collection is not converted. One can argue that this works as intended as the EventHeader collection is not listed there but I think we would always want to convert it as it is  not a normal collection in the lcio world.
https://github.com/key4hep/k4MarlinWrapper/blob/5e6bb7c6e59c974614bac91139f5107945a7f78f/test/gaudi_opts/fccRec_e4h_input.py#L78-L97

tmadlener:
Ah, now I get it. That makes sense. That should be fairly straight forward to fix by simply adding one line (plus a comment) before the `if` here: https://github.com/key4hep/k4MarlinWrapper/blob/5e6bb7c6e59c974614bac91139f5107945a7f78f/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L318-L320

Do you have any easy test that could also be added alongside with the fix? And do you have the time to implement the fix? (It will become end of this week at least for me).

Zehvogel:
Time for the fix maybe, easy test not so much. In principle we would need to check the lcio output to have consistent event numbers but as far as I can see that already requires multiple steps?

PR #162, opened 2023-11-13
Always convert EventHeader to lcio

Zehvogel:
BEGINRELEASENOTES
- Modified the edm4hep to lcio converter to always convert the`EventHeader` to have event number etc. available in lcio.

ENDRELEASENOTES

Does anyone have a good idea for a test?

Resolves: https://github.com/key4hep/k4MarlinWrapper/issues/153

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -88,6 +88,8 @@ if (BASH_PROGRAM)
   # multiple processors
   add_test( global_converter_maps ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/global_converter_maps.sh )
 
+  add_test( event_header ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/event_header.sh )
+
   set_tests_properties (
       simple_processors
       simple_processors2
@@ -100,6 +102,7 @@ if (BASH_PROGRAM)
       clicRec_edm4hep_input
       clic_geo_test
       global_converter_maps
+      event_header
     PROPERTIES
       ENVIRONMENT "TEST_DIR=${CMAKE_CURRENT_SOURCE_DIR};LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:${CMAKE_INSTALL_PREFIX}/lib64:$ENV{LD_LIBRARY_PATH};PYTHONPATH=${CMAKE_INSTALL_PREFIX}/python:$ENV{PYTHONPATH};EXAMPLE_DIR=${PROJECT_SOURCE_DIR}/k4MarlinWrapper/examples;MARLIN_DLL=$ENV{MARLIN_DLL}:${CMAKE_CURRENT_BINARY_DIR}/libMarlinTestProcessors.so"
       )
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -102,7 +102,7 @@ if (BASH_PROGRAM)
       clicRec_edm4hep_input
       clic_geo_test
       global_converter_maps
-      event_header
+      event_header_conversion
     PROPERTIES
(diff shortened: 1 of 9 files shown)


Zehvogel:
Unfortunately, I cannot reproduce the ci failure locally, for me all tests pass :/

tmadlener:
Is the `EventHeader` missing from here? 
https://github.com/key4hep/k4MarlinWrapper/blob/2eeb466f8af579fc970b6348b06076c8b52e09b8/test/gaudi_opts/test_global_converter_maps.py#L41

Zehvogel:
> Is the `EventHeader` missing from here?
> 
> https://github.com/key4hep/k4MarlinWrapper/blob/2eeb466f8af579fc970b6348b06076c8b52e09b8/test/gaudi_opts/test_global_converter_maps.py#L41

I guess so, lets see. Still puzzled why I don't get an error locally :D

tmadlener:
Maybe the better solution in this case would be to make sure that the `PodioInput` also reads the EventHeader collection?

Zehvogel:
hmpf yes I found why it does not error locally for me... the tests also need the install beforehand so if I click on run tests in my editor like a lazy person they run against the k4marlinwrapper from the environment e.g. nightlies

Zehvogel:
> Maybe the better solution in this case would be to make sure that the `PodioInput` also reads the EventHeader collection?

I agree

Zehvogel:
Waiting for
- [ ] https://github.com/key4hep/EDM4hep/pull/239 to be able to remove the plain string here
- [ ] https://github.com/key4hep/k4FWCore/pull/160 to be able to remove the explicit reading of the EventHeader

test/gaudi_opts/test_global_converter_maps.py
```diff
@@ -38,7 +38,7 @@
 )
 
 podioInput = PodioInput("InputReader")
-podioInput.collections = ["MCParticles"]
+podioInput.collections = ["MCParticles", "EventHeader"]
```
tmadlener, line 41: Do we still need the `EventHeader` after merging https://github.com/key4hep/k4FWCore/pull/160 ? If no, then I think we should remove it here.

Zehvogel: we should be able to remove it and that was my plan for today :)

tmadlener:
It seems we still need it: https://github.com/key4hep/k4MarlinWrapper/actions/runs/6887433198/job/18734625662?pr=162#step:6:2428

I am not entirely sure why though, we pick up the latest version of k4FWCore.

Zehvogel:
I think I know why, thought I tested it but as we set the collections to read after calling the constructor we just overwrite the list that has the EventHeaderName in it :(

Zehvogel:
so appending the collections works, not sure if this is really an improvement :/

Zehvogel:
I changed the steering files for the tests back again as https://github.com/key4hep/k4FWCore/pull/161 made the change unnecessary.

tmadlener (commented):
I would still like to have tests for this, but the "simplest" I can think of is to write a Marlin processor that we then run wrapped and check whether the contents of the EventHeader are as expected (first filling them with the `EventHeaderCreator` algorithm). The processor is almost trivial, but it introduces quite a bit of boiler plate...

Zehvogel:
`anajob` will print the real event numbers, so we could write out the lcio file and use a bit of ugly bash :D
```c++
        cout << "///////////////////////////////////" << endl;
        cout << "EVENT: " << evt->getEventNumber() << endl
            << "RUN: " << evt->getRunNumber() << endl
            << "DETECTOR: " << evt->getDetectorName() << endl
            << "COLLECTIONS: (see below)" << endl;
        cout << "///////////////////////////////////" << endl << endl;
```

test/CMakeLists.txt
```diff
@@ -88,6 +88,8 @@ if (BASH_PROGRAM)
   # multiple processors
   add_test( global_converter_maps ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/global_converter_maps.sh )
 
+  add_test( event_header ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/event_header.sh )
```
tmadlener, line 91:
```suggestion
  add_test( event_header_conversion ${BASH_PROGRAM} ${CMAKE_CURRENT_SOURCE_DIR}/scripts/event_header.sh )
```

For a bit more clarity when reading logs of CI.

Zehvogel:
@jmcarcell looks like the doctests are failing because `jupytext` is now missing from the stack?

PR #165, opened 2024-01-23
Don't use GaudiAlg

jmcarcell:
BEGINRELEASENOTES
- Change headers and add EventContext in algorithms not to use GaudiAlg
- Replace `GaudiTool` with `AlgTool`

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/CMakeLists.txt
+++ b/k4MarlinWrapper/CMakeLists.txt
@@ -26,7 +26,7 @@ gaudi_add_module(k4MarlinWrapperPlugins
     src/components/LcioEventAlgo.cpp
     src/components/LcioEventOutput.cpp
   LINK
-    Gaudi::GaudiAlgLib
+    Gaudi::GaudiKernel
     ${Marlin_LIBRARIES}
 )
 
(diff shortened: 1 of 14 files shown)


andresailer:
This is still work in progress or depend on some other PR?

jmcarcell:
This should now be ready, finally the build is passing

tmadlener (commented):
Do I have to understand why `AlgTool` is the non-deprecated on while `GaudiTool` is? ;)

This looks fairly complete now, or is there still something missing?

jmcarcell:
> Do I have to understand why `AlgTool` is the non-deprecated on while `GaudiTool` is? ;)
`GaudiTool` is in `GaudiAlg` :smile: 
I think there are a few differences like `AlgTool` not accessing the data store directly. In general I think we use few things of the class and we could probably get by with a minimal implementation of `IAlgTool`.

> This looks fairly complete now, or is there still something missing?
It's ready, I don't think anything needs to be changed unless we find issues but the tests are passing.

tmadlener (approved):

PR #166, opened 2024-01-27
Add the edm4hep includes to allow compilation with clang

jmcarcell:
BEGINRELEASENOTES
- Add the edm4hep includes to allow compilation with clang, which currently doesn't compile due to incomplete classes

ENDRELEASENOTES

Possible related discussion about different behavior for gcc and clang:
https://stackoverflow.com/questions/22770318/incomplete-types-in-template-code

With this change I can compile both on gcc and clang. Previously all the classes had to be specified anyway so having the includes is not that much of a change.

### DIFF
--- a/k4MarlinWrapper/src/components/GlobalConvertedObjectsMap.h
+++ b/k4MarlinWrapper/src/components/GlobalConvertedObjectsMap.h
@@ -21,20 +21,19 @@
 
 #include "k4EDM4hep2LcioConv/MappingUtils.h"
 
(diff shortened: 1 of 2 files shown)


tmadlener (commented):
I suppose it works for the LCIO types because they are used as pointers? Anyhow, no objections from my side here. We can then presumably remove the includes from the implementation file, right?

jmcarcell:
Yeah I think it's because of the pointers. I removed some includes, I'm not sure if I got them all

PR #172, opened 2024-02-07
Switch to alma9 container with CVMFS for doctests

tmadlener:
BEGINRELEASENOTES
- Switch the `doctest` CI workflows to an alma9 based container with CVMFS inside

ENDRELEASENOTES

Mainly doing this because the CentOS7 workflow is broken at the moment (https://github.com/key4hep/key4hep-spack/issues/558) and additionally, Alma9 is on lxplus by default now, so we should test on that rather than on the deprecated CentOS7.


.github/workflows/doctest.yaml
```diff
@@ -13,10 +13,9 @@ jobs:
                'doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper']
     steps:
     - uses: actions/checkout@v3
-    - uses: cvmfs-contrib/github-action-cvmfs@v3.1
     - name: Start container
       run: |
-        docker run -it --name CI_container -v ${GITHUB_WORKSPACE}:/Package -v /cvmfs:/cvmfs:shared -d ghcr.io/aidasoft/centos7:latest /bin/bash
+        docker run -it --name CI_container --privileged -v ${GITHUB_WORKSPACE}:/Package -v /cvmfs:/cvmfs:shared -d ghcr.io/key4hep/key4hep-images/alma9-cvmfs /bin/bash
```
vvolkl, line 18: Then I think you no longer need the bind mount
```suggestion
        docker run -it --name CI_container --privileged -v ${GITHUB_WORKSPACE}:/Package -d ghcr.io/key4hep/key4hep-images/alma9-cvmfs /bin/bash
```

.github/workflows/doctest.yaml
```diff
@@ -13,10 +13,9 @@ jobs:
...
     - name: Compile
       run: |
         docker exec CI_container /bin/bash -c ' cd Package;/
```
vvolkl, line 21: This also doesn't look right - has been broken for a while apparently :see_no_evil: 

```suggestion
        docker exec CI_container /bin/bash -c ' cd Package;\
```

tmadlener: It was broken for a while because we didn't build the necessary jupytext any longer. But that should have been fixed now.

tmadlener: Although this bit has definitely been working in the past

vvolkl: Yes, I guess the newline escape isn't needed, then it would just try to execute the `/`, fail, and continue.

.github/workflows/doctest.yaml
```diff
@@ -13,25 +13,27 @@ jobs:
...
     - name: CheckPage
       run: |
-        docker exec CI_container /bin/bash -c 'cd ./Package;
+        docker exec CI_container /bin/bash -c '/mount.sh
+        cd ./Package;
         source ${{ matrix.SETUP }};
         export LD_LIBRARY_PATH=/Package/install/lib64:$LD_LIBRARY_PATH
         export PYTHONPATH=/Package/install/python:$PYTHONPATH
         export K4MARLINWRAPPER=/Package/install/share/k4MarlinWrapper
+        unset KEY4HEP_STACK;
```
tmadlener, line 36: @jmcarcell this is rather brute-force but it should work. Should we think about a flag to the setup script that simply doesn't do anything if a key4hep stack is already setup?

tmadlener:
Merging this for now to unblock some other PRs.

PR #178, opened 2024-03-15
Modernize event display example steering

Zehvogel:
BEGINRELEASENOTES
- Modernize event display example

ENDRELEASENOTES

I modernized the event display example a bit and made it able to deal with both lcio and edm4hep files.

[The corresponding documentation](https://github.com/key4hep/k4MarlinWrapper/blob/main/doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md) still needs to be updated. However, it contains a lengthy step by step description of the steering file content that is no longer valid and the only thing I can offer regarding those parts is to delete them :shrug: but before doing so I wanted to gather some feedback :)

### DIFF
--- a/doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
+++ b/doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
@@ -16,9 +16,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-# Running an Event Display with EDM4hep input
+# Running an Event Display with EDM4hep or LCIO input
 
-It is possible to run the [C Event Display (CED)](https://github.com/iLCSoft/CED) via a wrapped [CEDViewer](https://github.com/iLCSoft/CEDViewer) Marlin Processor. This makes it possible to run the Event Dispaly with EDM4hep input files using an on the fly conversion to LCIO for CED. This introduction shows the basic concepts and also provides a options file that should work for most use cases. This example will be using the CLIC detector but should also work for other DD4hep detector models. The example is fully self contained, if you already have everything set up you can jump directly to [running the event display](#running-the-event-display).
+It is possible to run the [C Event Display (CED)](https://github.com/iLCSoft/CED) via a wrapped [CEDViewer](https://github.com/iLCSoft/CEDViewer) Marlin Processor. This makes it possible to run the Event Display with EDM4hep input files using an on the fly conversion to LCIO for CED. This introduction shows the basic concepts and also provides a options file that should work for most use cases. This example will be using the CLIC detector but should also work for other DD4hep detector models. The example is fully self contained, if you already have everything set up you can jump directly to [running the event display](#running-the-event-display).
 
 ## Setting up an environment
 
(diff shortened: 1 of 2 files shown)


k4MarlinWrapper/examples/event_display.py
```diff
@@ -18,58 +18,71 @@
...
+    # TODO: maybe modernize this
+    compact_file = "CLICPerformance/Visualisation/CLIC_o3_v06_CED/CLIC_o3_v06_CED.xml"
+
+geoSvc = GeoSvc("GeoSvc")
+geoSvc.detectors = [compact_file]
+geoSvc.OutputLevel = INFO
+geoSvc.EnableGeant4Geo = False
+svcList.append(geoSvc)
+
+def create_reader(input_files):
```
tmadlener, line 60: Since I wrote a very similar thing (at least in purpose) for ILDConfig, I think this could be a useful function to have importable from k4MarlinWrapper. Since we already have some python code for which we do that, this should be fairly easy to add there.

k4MarlinWrapper/examples/event_display.py
```diff
@@ -197,6 +211,7 @@
                               "PandoraPFOCollection", "0", "3", "7",
                               "JetOut", "0", "0", "3"],
                           "DrawSurfaces": ["true"],
+                          # "DrawSurfaces": ["false"],
```
tmadlener, line 214: Do I understand correctly, that we had this option in here twice with opposite values before?

k4MarlinWrapper/examples/event_display.py
```diff
@@ -18,58 +18,71 @@
...
+
+evtsvc = k4DataSvc("EventDataSvc")
+svcList.append(evtsvc)
+
+if reco_args.compactFile:
+    compact_file = reco_args.compactFile
+else:
+    # add the same default as before
+    # TODO: maybe modernize this
+    compact_file = "CLICPerformance/Visualisation/CLIC_o3_v06_CED/CLIC_o3_v06_CED.xml"
```
tmadlener, line 52: This could be handled via `default=<...>` in the corresponding `add_argument` call. Unless we make it a required argument and then `argparse` will error out for us if there is none. I actually like that solution best, because what good is a default geometry for an event display that is used by different communities?

tmadlener (commented):
Thanks. A few smallish comments inline.

As far as the README is concerned, I would do the following:
- Remove all the outdated parts that were describing the steps that were necessary to get from the automatically converted on to a working version related to handling of EDM4hep inputs.
- Remove the explanation of the detailed technical explanation of running things through the MarlinWrapper and rather reference other existing documentation for this
- Update the section of how this script was obtained originally (mainly rephrasing the last two bullet points to some general information about handling EDM4hep input and replacing `InitDD4hep` with the `GeoSvc`)

k4MarlinWrapper/examples/event_display.py
Zehvogel: Yours is not only very similar this is mostly an exact copy of it ;) 

I agree that we should have it importable, I also want to re-use it for the CLD reconstruction. Should I add it to this PR?

tmadlener: No real preference whether it is in this PR or another. I will let you decide.

k4MarlinWrapper/examples/event_display.py
Zehvogel: I am all for having to explicitly pass the detector model.

 I tried to keep some of the original behaviour when I wrote this but as I later on noticed that I need to rewrite the documentation anyway we can also change it now :)

k4MarlinWrapper/examples/event_display.py
```diff
@@ -18,58 +18,53 @@
...
+svcList.append(geoSvc)
 
-MyEventSelector = MarlinProcessorWrapper("MyEventSelector")
-MyEventSelector.OutputLevel = INFO
-MyEventSelector.ProcessorType = "EventSelector"
-MyEventSelector.Parameters = {
-                              "EventList": ["28", "0", "33", "0", "52", "0", "63", "0", "73", "0", "78", "0"]
-                              }
+
+if reco_args.inputFiles:
```
tmadlener, line 62: This should be impossible to be false with the `nargs="+"`, I think. On a slightly more "philosophical level", do we even want an event display without any events?

k4MarlinWrapper/examples/event_display.py
```diff
@@ -18,58 +18,53 @@
...
+geoSvc.OutputLevel = INFO
+geoSvc.EnableGeant4Geo = False
+svcList.append(geoSvc)
 
-MyEventSelector = MarlinProcessorWrapper("MyEventSelector")
-MyEventSelector.OutputLevel = INFO
-MyEventSelector.ProcessorType = "EventSelector"
-MyEventSelector.Parameters = {
-                              "EventList": ["28", "0", "33", "0", "52", "0", "63", "0", "73", "0", "78", "0"]
-                              }
```
tmadlener, line 72: This could be nice to have and should be fairly straight forward to do vi cl arg parsing? E.g.: https://github.com/AIDASoft/podio/blob/400f4f06bef3e7666c71e31be579ec715c412b8a/tools/podio-dump#L168-L175

tmadlener (approved):

k4MarlinWrapper/examples/event_display.py
Zehvogel: It is still possible :( but I guess I can make it a required argument.
(which gets us into a different philosophical discussion, but I am fine with doing it for this and for the compact file)

Zehvogel: urgh but it breaks `--help` then

tmadlener: Ah right, because we go through the whole `k4run` machinerie as well. Then having `--help` working is the better option.

k4MarlinWrapper/examples/event_display.py
Zehvogel: Getting this list as an argument yes. But I don't know how we would then select those?

We don't have an equivalent to the `<if>` construct in Marlin so we can not use the `EventSelector` processor. 
Is this possible with the `EvtSel` of the `ApplicationMgr`?
Or is there any other (already implemented) way?

tmadlener: Ah right, I wasn't aware that the `EventSelector` uses the `if` machinery. Then my comment is moot and we will have to find another way to do this.

tmadlener (commented):
I think the only thing left to do would be to make the compactFile a command line argument. After that I would merge this (with the EventSelector removed).

Zehvogel:
done (I hope)

doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
```diff
@@ -53,87 +53,35 @@ You should now have a `gamma_10GeV_edm4hep.root` file containing 10 events.
...
-- Using the `convertMarlinSteeringToGaudi.py` converter script convert this into a Gaudi options file
-- Exchange the LCIO input reading by the podio input reading (see above)
-- Attach the `EDM4hep2LcioTool` to the wrapped `CEDViewer` processor
+When running this for a detector different than CLD, CLIC or ILD (or a derivative of one of them) you might not see anything in the event display.
 
-This should allow one to arrive at a similar steering file even for slightly different configurations. One potential pitfall is the slightly different naming of the `ddsim` outputs between LCIO and EDM4hep (see [this issue](https://github.com/AIDASoft/DD4hep/issues/921)). This can be addressed by configuring the EDM4hep to LCIO converter (`edmConvTool` in the code above) to map the names accordingly.
+If you are not seeing the detector make sure that it has the proper visualisation attributes, e.g. by comparing it to the compact file used above.
+If you are not seeing any hits, tracks etc. from your event, make sure that to add their collection names to the `DrawInLayer` list in the `DDCEDViewer` parameters.
```
tmadlener, line 87:
```suggestion
If you are not seeing any hits, tracks etc. from your event, make sure to add their collection names to the `DrawInLayer` list in the `DDCEDViewer` parameters.
```

tmadlener (commented):
Apart from the typo, I think this is good to go.

doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
Zehvogel: thx!

doc/starterkit/k4MarlinWrapperCLIC/CEDViaWrapper.md
```diff
@@ -53,87 +53,35 @@ You should now have a `gamma_10GeV_edm4hep.root` file containing 10 events.
...
-```
-Note that in this case we already have passed in the geometry that we want to use for the event display.
+## Details
 
-Finally, the main work is done by the `DDCEDViewer`, which we again use via the `MarlinProcessorWrapper` (omitting a lot of the detailed configuration here). In order to convert the EDM4hep inputs to LCIO we attach an `EDM4hep2LcioTool` to this algorithm
-```python
-from Configurables import EDM4hep2LcioTool
+The main work is done by the `DDCEDViewer`, which we use via the `MarlinProcessorWrapper`. It is the following part of example `event_display.py`.
 
+```python
```
tmadlener, line 72:
```suggestion
```python
from Configurables import MarlinProcessorWrapper
```
To make the doctest happy. Potentially you also have to remove the `OutputLevel` (or import `INFO` from `Gaudi.Configuration`)

tmadlener:
Thanks. Failing CI is now only due to known issues.

PR #179, opened 2024-03-18
Add inputReader for easy reading of edm4hep or lcio files

Zehvogel:
BEGINRELEASENOTES
- Add inputReader for easy reading of edm4hep or lcio files

ENDRELEASENOTES

see discussion in #178

### DIFF
--- a/k4MarlinWrapper/python/k4MarlinWrapper/inputReader.py
+++ b/k4MarlinWrapper/python/k4MarlinWrapper/inputReader.py
@@ -18,7 +18,7 @@
 #
 
 import sys
-from Configurables import LcioEvent, PodioInput
+from Configurables import LcioEvent, PodioInput, MarlinProcessorWrapper, EDM4hep2LcioTool
 
 def create_reader(input_files, evtSvc):
     """Create the appropriate reader for the input files"""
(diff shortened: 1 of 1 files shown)


tmadlener:
Since we are on the topic of making this "easy to use". Is there an easy way to also deal with the conditional attaching of a converter? From a quick look at the code we would need the `algList` and attach it to the first `MarlinWrapper` algorithm? That could also check that the first / only algorithm before that is the `PodioInput` (or do nothing in case it's an `LcioEvent`?

Maybe there is something more clever, but I would be happy with a two step/function approach for the moment.

Zehvogel:
I will add conversion in the other direction in another pr later

k4MarlinWrapper/python/k4MarlinWrapper/inputReader.py
```diff
@@ -0,0 +1,56 @@
...
+        if any(not f.endswith(".root") for f in input_files):
+            print("All input files need to have the same format (EDM4hep)")
+            sys.exit(1)
+        read = PodioInput("PodioInput")
+        evtSvc.inputs = input_files
+
+    return read
+
+
+def attach_edm4hep2lcio_conversion(algList, read):
```
tmadlener, line 42: Can you add a brief description as a docstring that contains the fact that this is pretty much what you want to do if you have used `create_reader` unconditionally?

tmadlener (commented):
Apart from a bit of missing documentation I am very happy with this. Thanks :)

k4MarlinWrapper/python/k4MarlinWrapper/inputReader.py
Zehvogel: done :)

k4MarlinWrapper/python/k4MarlinWrapper/inputReader.py
```diff
@@ -0,0 +1,56 @@
...
+    """Attach the edm4hep to lcio conversion if necessary e.g. when using create_reader. Should only be run after algList is complete."""
+    if not isinstance(read, PodioInput):
+        # nothing to convert :)
+        return
+
+     # find first wrapper
+    for alg in algList:
+        if (isinstance(alg, MarlinProcessorWrapper)):
```
jmcarcell, line 49:
```suggestion
        if isinstance(alg, MarlinProcessorWrapper):
```

Zehvogel: thx

tmadlener (approved):

Issue #175, opened 2024-02-20
Set Marlin processor Verbosity from `MarlinProcessorWrapper.OutputLevel`


Zehvogel:
This is a bit confusing at the moment as the log level for the `MarlinProcessorWrapper` Gaudi algorithm and the run Marlin processor can be inconsistent. It will require some kind of mapping and it might not be 1:1 in all cases, but it might lessen user confusion. At the same time we could also warn about the necessity of `MARLINDEBUG=1`.

https://github.com/key4hep/k4MarlinWrapper/blob/0735c0dbb671f7ec8d16fbcf4f5cb02c1c5afe60/k4MarlinWrapper/src/components/MarlinProcessorWrapper.cpp#L111-L114

PR #182, opened 2024-04-17
Harmonize verbosities

Zehvogel:
BEGINRELEASENOTES
- By default use the `MarlinProcessorWrapper.OutputLevel` to set the verbosity of the wrapped Marlin processor.

ENDRELEASENOTES

It is still possible to overwrite the verbosity from the wrapped parameters. I also changed the default to `ERROR` (least output) this then also behaves as expected when even stricter Gaudi msg levels are choosen like `FATAL` or `ALWAYS`.

fixes: #175


andresailer (approved):
Thanks!

PR #167, opened 2024-01-28
Use ExternalData and remove some scripts for testing

jmcarcell:
BEGINRELEASENOTES
- Use ExternalData and remove some scripts for testing to fix running tests in parallel

ENDRELEASENOTES

Tests are failing sometimes; there are different scripts trying to download the same files and there are some race conditions. With these changes the files are downloaded at configure time as we do with other repositories.

PR #183, opened 2024-04-24
Add missing `-`

Zehvogel:
Looks like this went missing in #167?

BEGINRELEASENOTES
- Reintroduce missing minus sign in test over all events

ENDRELEASENOTES


tmadlener (approved):
good catch.

PR #184, opened 2024-05-07
Convert event parameters from LCIO to EDM4hep

Zehvogel:
BEGINRELEASENOTES
- Add lcio to edm4hep event parameter conversion

ENDRELEASENOTES

- [x] depends-on: https://github.com/key4hep/k4FWCore/pull/194

### DIFF
--- a/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
+++ b/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
@@ -130,6 +130,11 @@ namespace {
 }  // namespace
 
 StatusCode Lcio2EDM4hepTool::convertCollections(lcio::LCEventImpl* the_event) {
+  // Convert event parameters
+  // auto& frame = const_cast<podio::Frame&>(m_podioDataSvc->getEventFrame());
+  auto& frame = m_podioDataSvc->m_eventframe;
+  LCIO2EDM4hepConv::convertObjectParameters<lcio::LCEventImpl>(the_event, frame);
+
   // Convert Event Header outside the collections loop
   if (!collectionExist(edm4hep::EventHeaderName)) {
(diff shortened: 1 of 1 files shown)


tmadlener:
Technically, I think the `const_cast` should be OK in this case as the original Frame is not `const`. Additionally, the underlying `putParameter` (and the `GenericParameters`) are thread safe (via locking). From that point of view the `PodioDataSvc` could in principle also hand out a mutable Frame via `getEventFrame`.

Zehvogel:
Yes I mainly meant it has to go as I would consider it ugly...

I am in favour of returning a non-const frame to get rid of this quickly. :)

I am not so sure about the sustainability in the long run though. The whole being able to access the frame directly via the datasvc from any Algorithm leads the whole concept of Algorithms having to specify in-/output collections ad absurdum, doesn't it?

tmadlener:
> I am not so sure about the sustainability in the long run though. The whole being able to access the frame directly via the datasvc from any Algorithm leads the whole concept of Algorithms having to specify in-/output collections ad absurdum, doesn't it?

Yes, fully agree. I am not sure if there is an easier way at the moment. In principle I think only the MarlinWrapper currently needs these special accesses. But that will probably also be around for quite some time.

Zehvogel:
can someone re-trigger the ci pls? :)

andresailer:
Can you not?

Zehvogel:
> Can you not?

No, and I think it is not only because I am too blind to find the button...

andresailer:
> > Can you not?
> 
> No, and I think it is not only because I am too blind to find the button...

True, https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs

But the jobs still fail.

Zehvogel:
Oh yes, the merge this morning just felt so far away that I thought it was yesterday... it will take another day until the nightlies picked it up :(

tmadlener (approved):

PR #187, opened 2024-06-19
Fix a few warnings when compiling

jmcarcell:
BEGINRELEASENOTES
- Fix warnings after https://github.com/key4hep/EDM4hep/pull/315
- Fix a warning by adding a default case in a switch statement

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -93,7 +93,8 @@ void EDM4hep2LcioTool::convertTrackerHits(TrackerHitMap& trackerhits_vec, const 
   DataHandle<edm4hep::TrackerHit3DCollection> trackerhits_handle{e4h_coll_name, Gaudi::DataHandle::Reader, this};
   const auto                                  trackerhits_coll = trackerhits_handle.get();
 
-  MetaDataHandle<std::string> cellIDStrHandle{trackerhits_handle, edm4hep::labels::CellIDEncoding, Gaudi::DataHandle::Reader};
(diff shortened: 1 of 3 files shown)


tmadlener (approved):
Is this currently warning free? Then we could  consider adding `-Werror` to CI to keep it that way? But I suppose that would need to go into the key4hep build action, right?

jmcarcell:
I don't get any warnings with this change. Adding `-Werror` won't work for `k4EDM4hep2LcioConv` so I can't add it globally yet but I'm not opposed to have it or at least try it for some time. It may be annoying when changing compiler or version and finding that a warning appears though.

tmadlener:
Maybe it could be an option of the action? Then it would be easier to toggle on a per project basis. Maybe we need to take the discussion to the correct repo. I will merge this now.

jmcarcell:
Tagging @Zehvogel since he has complained about message verbosity; this change will modify the cases when the Gaudi message level is set to `NIL, VERBOSE, FATAL, ALWAYS, NUM_LEVELS` (see https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/install/include/GaudiKernel/IMessageSvc.h#L25), since these were not in the switch statement. At the very least `m_level` is set by default to `NIL` https://gitlab.cern.ch/gaudi/Gaudi/-/blob/master/install/include/GaudiKernel/CommonMessaging.h#L130. I think this may change the default when no verbosity is set since it's `ERROR` https://github.com/key4hep/k4MarlinWrapper/blob/main/k4MarlinWrapper/k4MarlinWrapper/MarlinProcessorWrapper.h#L68 and after this it will be `MESSAGE`.

PR #188, opened 2024-06-24
Add support for running together with functional algorithms

jmcarcell:
These are the minimal changes needed to make the wrapper work with functional algorithms using `PodioInput` and `PodioOutput`. Working with the new `IOSvc` will take a bit more work since more changes are needed.

BEGINRELEASENOTES
- Add support for running together with functional algorithms
- Add a test that creates MCParticles in EDM4hep with a functional algorithm that are later used by Marlin processors

ENDRELEASENOTES


k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -284,9 +284,19 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
   if (!collPtr) {
-    error() << "No collection with name: " << e4h_coll_name << " available for conversion" << endmsg;
+    DataObject* p;
+    auto        sc  = m_podioDataSvc->retrieveObject(e4h_coll_name, p);
+    auto        ptr = dynamic_cast<AnyDataWrapper<std::shared_ptr<podio::CollectionBase>>*>(p);
+    if (sc.isFailure() || !ptr) {
+      error() << "No collection with name: " << e4h_coll_name << " available for conversion" << endmsg;
+    } else {
+      collPtr = dynamic_cast<podio::CollectionBase*>(ptr->getData().get());
+    }
```
tmadlener, line 299: Is this path not possible for cases where it is in a Frame? I.e. could we just have one path through this, and drop the usage of the `evtFrame`?

tmadlener:
Looking at the test that you are hooking on here, I think we could even remove the current `PseudoRecoProcessor` and instead replace it with this algorithm and update the corresponding README. That would mean that we don't have effectively duplicated tests with slightly different implementations (and less code to maintain in the long run).

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
jmcarcell: I think this question mixes two things: the collection is either in the Frame if it has been produced with the non-functional algorithms or it is not if it has been produced by functional algorithms; then we can not use the event frame by asking the data service for the collections instead (but this could have been done earlier and doesn't have much to do with the current changes) as it is right now.

jmcarcell:
`PseudoRecoProcessor` removed.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
tmadlener: I was more aiming at the question, whether we can reach collections that have been produced by non-functional algorithms via the data svc, instead of going to the event Frame?

jmcarcell:
Any more comments or can we merge this?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -282,11 +282,27 @@ void EDM4hep2LcioTool::convertEventHeader(const std::string& e4h_coll_name, lcio
...
+  podio::CollectionBase* collPtr  = nullptr;
+  DataObject*            p;
+  auto                   sc = m_podioDataSvc->retrieveObject(e4h_coll_name, p);
+  if (sc.isFailure()) {
+    throw GaudiException("Collection not found", name(), StatusCode::FAILURE);
+  }
+  auto ptr = dynamic_cast<DataWrapperBase*>(p);
+  if (ptr) {
+    collPtr = ptr->collectionBase();
+  }
```
tmadlener, line 295: Slightly more verbose than before, but at least we got the detail of using a Frame out of here.

tmadlener (approved):
Thanks. Looks good to me.

PR #200, opened 2024-08-23
Switch from association to link

### LINKED ISSUE #199: Fix Association checks
https://github.com/key4hep/k4MarlinWrapper/blob/b6c0ec6b74990fd40db276e923d54e06cee5fb54/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L388

tmadlener:
BEGINRELEASENOTES
- Switch from `Association` to `Link` in the conversion from EDM4hep to LCIO to accomodate for the new naming from ([EDM4hep#341](https://github.com/key4hep/EDM4hep/pull/341))
- Add test to make sure that conversion of Links from EDM4hep to LCIO works as expected

ENDRELEASENOTES

Fixes #199 (I think). @andresailer was this what you were referring to in that issue?
Fixes https://github.com/key4hep/k4EDM4hep2LcioConv/issues/93

### DIFF
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -381,13 +381,13 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
   std::vector<EDM4hep2LCIOConv::ParticleIDConvData> pidCollections{};
   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>  dQdxCollections{};
 
-  std::vector<std::tuple<std::string, const podio::CollectionBase*>> associations{};
+  std::vector<std::tuple<std::string, const podio::CollectionBase*>> linkCollections{};
 
   for (const auto& [edm4hepName, lcioName] : collsToConvert) {
     const auto coll = getEDM4hepCollection(edm4hepName);
-    if (coll->getTypeName().find("Association") != std::string_view::npos) {
-      debug() << edm4hepName << " is an association collection, converting it later" << endmsg;
-      associations.emplace_back(lcioName, coll);
+    if (coll->getTypeName().find("LinkCollection") != std::string_view::npos) {
+      debug() << edm4hepName << " is a link collection, converting it later" << endmsg;
+      linkCollections.emplace_back(lcioName, coll);
       continue;
     }
     debug() << "Converting collection " << edm4hepName << " (storing it as " << lcioName << ")" << endmsg;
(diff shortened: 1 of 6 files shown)


k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -381,13 +381,13 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
   std::vector<EDM4hep2LCIOConv::ParticleIDConvData> pidCollections{};
   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>  dQdxCollections{};
 
-  std::vector<std::tuple<std::string, const podio::CollectionBase*>> associations{};
+  std::vector<std::tuple<std::string, const podio::CollectionBase*>> linkCollections{};
 
   for (const auto& [edm4hepName, lcioName] : collsToConvert) {
     const auto coll = getEDM4hepCollection(edm4hepName);
-    if (coll->getTypeName().find("Association") != std::string_view::npos) {
```
jmcarcell, line 388: This is what #199 is referring to since now it's impossible to find "Association" in the type name.

tmadlener:
Seems, I didn't merge this yet. AFAIU this solves the issue and can be merged, right?

jmcarcell:
Maybe there should be a test doing something with links? Otherwise this can just not be caught if there is another problem like this one.

tmadlener:
There is already a test with links [`global_converter_maps`](https://github.com/key4hep/k4MarlinWrapper/blob/main/test/README.md#global_converter_maps) that goes from LCIO to EDM4hep (which works without problems apparently). I added a new test that also checks the other direction. I added this as a completely separate test that uses one new `Transformer` for creating the `Links` in EDM4hep and then it uses another new Marlin `Processor` for checking it's contents. Prior to the fixes this test fails, so it does indeed check that links are converted.

test/src/MarlinMCRecoLinkChecker.cc
```diff
@@ -0,0 +1,89 @@
...
+
+class MarlinMCRecoLinkChecker : public marlin::Processor {
+public:
+  MarlinMCRecoLinkChecker();
+
+  marlin::Processor* newProcessor() override { return new MarlinMCRecoLinkChecker; }
+
+  /** process the event - In this case simply link MCparticle[i] with
+   * ReconstructedParticle[i]
+   */
```
tmadlener, line 38:
```suggestion
```

tmadlener:
@jmcarcell do you want to have another closer look? Otherwise, I would merge this later today.

jmcarcell (approved):

test/src/MarlinMCRecoLinkChecker.cc
```diff
@@ -0,0 +1,86 @@
...
+    const auto mc   = static_cast<EVENT::MCParticle*>(mcColl->getElementAt(i));
+    const auto reco = static_cast<EVENT::ReconstructedParticle*>(recoColl->getElementAt(i));
+    const auto rel  = static_cast<EVENT::LCRelation*>(relColl->getElementAt(i));
+
+    if (rel->getWeight() != i) {
+      throw std::runtime_error("Relation " + std::to_string(i) + " does not have the correct weight (expected: " +
+                               std::to_string(i) + ", actual: " + std::to_string(rel->getWeight()) + ")");
+    }
+
+    if (!(rel->getTo() == mc)) {
```
jmcarcell, line 76: So we don't have `!=`?

tmadlener: We do, especially because for LCIO this is just comparing pointers. (We would also have it for EDM4hep). This is a remnant from copy-pasting.

PR #204, opened 2024-10-28
Retrieve unique_ptrs for collections produced by functional algorithms

jmcarcell:
BEGINRELEASENOTES
- Retrieve unique_ptrs for collections produced by functional algorithms. Needed after https://github.com/key4hep/k4FWCore/pull/250. It should also be able to handle the case for the previous approach with `std::shared_ptr`.

ENDRELEASENOTES


jmcarcell:
Merging so that it's available in the nightlies and for tagging tomorrow.

PR #205, opened 2024-12-06
Only add an OutputLevel if Verbosity is set

tmadlener:
Unlike Marlin, Gaudi doesn't override OutputLevels of Algorithms if they have been manually set in the options file. Hence, if we populate all wrapped processor OutputLevels it will require changing all of them manually afterwards if we want the global OutputLevel that is set in the AlgorithmMgr to apply.

BEGINRELEASENOTES
- Only set the `OutputLevel` for converted MarlinProcessors if `Verbosity` was set in the original XML steering file

ENDRELEASENOTES


jmcarcell (approved):
I think it makes sense and will make it easier for the verbosity to be controlled from Gaudi.

PR #207, opened 2024-12-09
Multiple minor fixes

jmcarcell:
BEGINRELEASENOTES
- Don't use the deprecated service retrieval
- Add `final` when possible and use the default destructors instead of implementing them
- Use `!` instead of `not` in `if(...)`
- Use `!` instead of comparing to `nullptr`
- Simplify setting the environment for tests, avoiding having to duplicate the names
- Clean up some includes in header and source files

ENDRELEASENOTES

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
@@ -26,19 +26,17 @@
  */
 
 #include <Gaudi/Algorithm.h>
-#include <GaudiKernel/IEventProcessor.h>
+#include <Gaudi/Property.h>
 
-#include <EVENT/LCIO.h>
-#include <MT/LCReader.h>
-
-#include "k4MarlinWrapper/LCEventWrapper.h"
+namespace MT {
+  class LCReader;
+}
 
 class LcioEvent : public Gaudi::Algorithm {
 public:
   explicit LcioEvent(const std::string& name, ISvcLocator* pSL);
-  virtual ~LcioEvent() = default;
-  virtual StatusCode initialize() override final;
-  virtual StatusCode execute(const EventContext&) const override;
+  StatusCode initialize() override final;
+  StatusCode execute(const EventContext&) const override;
 
 private:
   Gaudi::Property<std::vector<std::string>> m_fileNames{this, "Files", {}};
--- a/k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
@@ -35,13 +35,13 @@ namespace MT {
 class LcioEvent : public Gaudi::Algorithm {
 public:
   explicit LcioEvent(const std::string& name, ISvcLocator* pSL);
-  StatusCode initialize() override final;
-  StatusCode execute(const EventContext&) const override;
+  StatusCode initialize() final;
+  StatusCode execute(const EventContext&) const final;
 
 private:
   Gaudi::Property<std::vector<std::string>> m_fileNames{this, "Files", {}};
   MT::LCReader*                             m_reader = nullptr;
-  bool                                      isReEntrant() const override { return false; }
+  bool                                      isReEntrant() const final { return false; }
 };
 
 #endif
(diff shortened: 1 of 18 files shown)


k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -58,11 +53,11 @@ struct CollectionPairMappings;
 class EDM4hep2LcioTool : public AlgTool, virtual public IEDMConverter {
 public:
   EDM4hep2LcioTool(const std::string& type, const std::string& name, const IInterface* parent);
-  virtual ~EDM4hep2LcioTool();
-  virtual StatusCode initialize();
-  virtual StatusCode finalize();
+  ~EDM4hep2LcioTool() override = default;
```
m-fila, line 56:
```suggestion
```
No need to declare a destructor just override it to default - 'the rule of zero' recommends relying on the compiler-generated one
Same in `Lcio2EDM4hepTool`

m-fila (commented):
There are a few more destructors that could be cleaned-up (or removed according to rule of 0):
- `~LcioEvent`
- `~LCEventWrapper`
-  `~LcioEventOutput`
- `~MarlinProcessorWrapper`
- `~TrackingCellIDEncodingSvc`

k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
```diff
@@ -46,7 +46,7 @@ namespace EVENT {
 class Lcio2EDM4hepTool : public AlgTool, virtual public IEDMConverter {
 public:
   Lcio2EDM4hepTool(const std::string& type, const std::string& name, const IInterface* parent);
-  virtual ~Lcio2EDM4hepTool();
+  ~Lcio2EDM4hepTool() override = default;
```
m-fila, line 49:
```suggestion
```
I think this is the last remaining redundant destructor

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
jmcarcell: Done

k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
jmcarcell: Done

m-fila: Thanks :smiley:

andresailer:
For my own elucidation: why not `not`?

jmcarcell:
By convention, we use `!` almost everywhere.

k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
```diff
@@ -26,19 +26,17 @@
...
+}
 
 class LcioEvent : public Gaudi::Algorithm {
 public:
   explicit LcioEvent(const std::string& name, ISvcLocator* pSL);
-  virtual ~LcioEvent() = default;
-  virtual StatusCode initialize() override final;
-  virtual StatusCode execute(const EventContext&) const override;
+  StatusCode initialize() override final;
+  StatusCode execute(const EventContext&) const override;
```
tmadlener, line 39:
```suggestion
  StatusCode execute(const EventContext&) const override final;
```
This seems to be what we go for in all other places, even though `final` technically already implies `override` so that that is superfluous.

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -58,11 +53,10 @@ struct CollectionPairMappings;
 class EDM4hep2LcioTool : public AlgTool, virtual public IEDMConverter {
 public:
   EDM4hep2LcioTool(const std::string& type, const std::string& name, const IInterface* parent);
-  virtual ~EDM4hep2LcioTool();
-  virtual StatusCode initialize();
-  virtual StatusCode finalize();
+  StatusCode initialize() override;
+  StatusCode finalize() override;
```
tmadlener, line 57:
```suggestion
  StatusCode initialize() final;
  StatusCode finalize() final;
```
Simply because we do that for the other direction.

tmadlener (commented):
We are not really consistent with `override` and `final` here. Sometimes we use both, sometimes only one. I have a few examples below.

tmadlener (approved):

PR #209, opened 2024-12-09
Remove the check for TrackerHit3D from edm4hep

jmcarcell:
BEGINRELEASENOTES
- Remove the check for TrackerHit3D from edm4hep

ENDRELEASENOTES

See https://github.com/key4hep/k4FWCore/pull/265


tmadlener (approved):
We already require EDM4hep v0.99 in any case in the top level CMakeLists.txt

Issue #208, opened 2024-12-09
MarlinProcessorWrapper loops extra event with EvtMax=-1 in ApplicationMgr


dudarboh:
## Issue

Running Gaudi over all events with `EvtMax=-1` (the default from the [documentation](https://gaudi.web.cern.ch/doxygen/master/dc/d91/class_application_mgr.html#a7e707f5bf8b65cb667146fa677122b60)) tries to run one extra event with `MarlinProcessprWrapper`. Running with `EvtMax=manually_set_max_events` works without errors.

## Reproducible

Running the default reconstruction test from [ILDConfig](https://github.com/iLCSoft/ILDConfig)
```
k4run ILDReconstruction.py --inputFiles=bbudsc_3evt_SIM.slcio
```
with
```
EvtMax=-1
````

Relevant output in the end:

```
[ MESSAGE0 "MyRecoMCTruthLinker"]  processEvent 2  - 0
MyAIDAProcessor   WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
MyStatusmonitor   WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
MySplitCollecti...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
VXDPlanarDigiPr...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
SITPlanarDigiPr...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
FTDPixelPlanarD...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
FTDStripPlanarD...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
FTDDDSpacePoint...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
SETPlanarDigiPr...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
SETDDSpacePoint...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
MyTPCDigiProcessorWARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
MyClupatraProce...WARNING LCIO Event reading returned nullptr, so MarlinProcessorWrapper won't execute
[...]

PR #210, opened 2024-12-10
Make sure that reading of LCIO events stops early enough

tmadlener:
BEGINRELEASENOTES
- Stop the reading of the LCIO events early enough, such that no empty event is processed after the last event (which breaks "regular" Algorithms in the same chain)
- Make it a hard error (exception) when reading an event fails

ENDRELEASENOTES

This is iteration N of fixing #208

Some more things to consider in this whole thing:
- `marlin::SkipEventException` will most likely lead to problems if the chain has non wrapped algorithms
  - Gaudi technically has a similar thing with *revoverable* exceptions, but I am not entirely sure if it still works
- The `LCEventStatusWrapper` should probably renamed now that it is completely decoupled from the LCIO Event status.


k4MarlinWrapper/k4MarlinWrapper/LcioEventAlgo.h
```diff
@@ -41,6 +41,8 @@ class LcioEvent : public Gaudi::Algorithm {
 private:
   Gaudi::Property<std::vector<std::string>> m_fileNames{this, "Files", {}};
   MT::LCReader*                             m_reader = nullptr;
+  int                                       m_numberOfEvents{};
+  mutable int                               m_currentEvent{};
```
tmadlener, line 45:
```suggestion
  mutable int                               m_currentEvent{};  // No atomicity necessary since not re-entrant
```

PR #216, opened 2025-01-14
Add support for running with IOSvc

### LINKED ISSUE #202: Converters do not work with `IOSvc`
The converters still partially refer to the `PodioDataSvc` explicitly and will fail with a segmentation violation when used with the `IOSvc`.

jmcarcell:
BEGINRELEASENOTES
- Add support for running the wrapper with `IOSvc` with a few tests:
  - Run the CLIC reconstruction with `IOSvc` (proves processors run under IOSvc, producing the same output).
  - Run `global_converter_maps` with `IOSvc`, where a functional algorithm produces output that is used by a Marlin processor and the output of the two is used either by a `Gaudi::Algorithm` or a functional algorithm (proves processors can take input from functional algorithms and functional algorithms can take input from processors).
  - Run `test_link_conversion_edm4hep` with `IOSvc`, where links produced by functional algorithms are checked by Marlin processors.

ENDRELEASENOTES

- [x] Needs https://github.com/key4hep/k4FWCore/pull/273
- [x] Needs https://github.com/key4hep/k4FWCore/pull/280

Fix #202

### DIFF
--- a/k4MarlinWrapper/CMakeLists.txt
+++ b/k4MarlinWrapper/CMakeLists.txt
@@ -54,6 +54,13 @@ target_include_directories(MarlinWrapper PUBLIC
   ${LCIO_INCLUDE_DIRS}
 )
 
+# k4MarlinWrapperUtils
+gaudi_add_library(k4MarlinWrapperUtils SHARED
+  SOURCES
+    src/components/StoreUtils.cpp
+  LINK k4FWCore::k4FWCore
+)
+
 # EDM4hep2lcio
 gaudi_add_module(EDM4hep2Lcio
   SOURCES
@@ -64,6 +71,7 @@ gaudi_add_module(EDM4hep2Lcio
     ${LCIO_LIBRARIES}
     ${Marlin_LIBRARIES}
     EDM4HEP::edm4hep
+    k4MarlinWrapperUtils
 )
 
 target_include_directories(EDM4hep2Lcio PUBLIC
@@ -80,6 +88,7 @@ gaudi_add_module(Lcio2EDM4hep
     EDM4HEP::edm4hep
     k4FWCore::k4FWCore
     k4EDM4hep2LcioConv::k4EDM4hep2LcioConv
+    k4MarlinWrapperUtils
 )
 
 target_include_directories(Lcio2EDM4hep PUBLIC
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -21,8 +21,6 @@
 
 #include "k4MarlinWrapper/converters/IEDMConverter.h"
 
-#include "k4FWCore/IMetadataSvc.h"
-
 #include "k4EDM4hep2LcioConv/k4EDM4hep2LcioConv.h"
 
 #include <Gaudi/Property.h>
@@ -33,7 +31,8 @@
 #include <vector>
 
 class PodioDataSvc;
-class MetadataSvc;
+class IDataProviderSvc;
+class IMetadataSvc;
 
 template <typename K, typename V> using ObjMapT = k4EDM4hep2LcioConv::VecMapT<K, V>;
 
@@ -66,7 +65,9 @@ private:
   Gaudi::Property<bool>                               m_convertAll{this, "convertAll", true};
 
   PodioDataSvc*                   m_podioDataSvc;
+  // EventDataSvc that is used together with IOSvc
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
+  // Metadata service from k4FWCore that is used together with IOSvc
   SmartIF<IMetadataSvc>           m_metadataSvc;
   std::vector<std::string>        m_collectionNames;
   std::map<uint32_t, std::string> m_idToName;
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -116,8 +116,6 @@ private:
                   std::vector<EDM4hep2LCIOConv::ParticleIDConvData>& pidCollections,
                   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>&  dQdxCollections);
 
-  StatusCode getAvailableCollectionsFromStore();
-
   /// Get an EDM4hep collection by name, consulting either the podio based data
   /// svc or the IOSvc
   podio::CollectionBase* getEDM4hepCollection(const std::string& name) const;
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -21,6 +21,8 @@
 
 #include "k4MarlinWrapper/converters/IEDMConverter.h"
 
+#include "k4FWCore/IMetadataSvc.h"
+
 #include "k4EDM4hep2LcioConv/k4EDM4hep2LcioConv.h"
 
 #include <Gaudi/Property.h>
@@ -31,6 +33,7 @@
 #include <vector>
 
 class PodioDataSvc;
+class MetadataSvc;
 
 template <typename K, typename V> using ObjMapT = k4EDM4hep2LcioConv::VecMapT<K, V>;
 
@@ -64,6 +67,7 @@ private:
 
   PodioDataSvc*                   m_podioDataSvc;
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
+  SmartIF<IMetadataSvc>           m_metadataSvc;
   std::vector<std::string>        m_collectionNames;
   std::map<uint32_t, std::string> m_idToName;
 
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -64,7 +64,7 @@ private:
   Gaudi::Property<std::map<std::string, std::string>> m_collNames{this, "collNameMapping", {}};
   Gaudi::Property<bool>                               m_convertAll{this, "convertAll", true};
 
-  PodioDataSvc*                   m_podioDataSvc;
+  PodioDataSvc* m_podioDataSvc;
   // EventDataSvc that is used together with IOSvc
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
   // Metadata service from k4FWCore that is used together with IOSvc
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -64,6 +64,8 @@ private:
 
   PodioDataSvc*                   m_podioDataSvc;
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
+  std::vector<std::string>        m_collectionNames;
+  std::map<uint32_t, std::string> m_idToName;
 
   void convertTracks(TrackMap& tracks_vec, const std::string& e4h_coll_name, const std::string& lcio_coll_name,
                      lcio::LCEventImpl* lcio_event);
@@ -109,6 +111,8 @@ private:
                   CollectionPairMappings&                            collection_pairs,
                   std::vector<EDM4hep2LCIOConv::ParticleIDConvData>& pidCollections,
                   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>&  dQdxCollections);
+
+  StatusCode getAvailableCollectionsFromStore();
 
   /// Get an EDM4hep collection by name, consulting either the podio based data
   /// svc or the IOSvc
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -73,7 +73,7 @@ StatusCode EDM4hep2LcioTool::initialize() {
   m_podioDataSvc = dynamic_cast<PodioDataSvc*>(m_eventDataSvc.get());
 
   m_metadataSvc = service("MetadataSvc", false);
-  if (!m_metadataSvc) {
+  if (!m_podioDataSvc && !m_metadataSvc) {
     error() << "Could not retrieve MetadataSvc" << endmsg;
     return StatusCode::FAILURE;
   }
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -34,7 +34,6 @@
 #include "GaudiKernel/AnyDataWrapper.h"
 #include "GaudiKernel/IDataManagerSvc.h"
 #include "GaudiKernel/IDataProviderSvc.h"
-#include "GaudiKernel/SmartDataPtr.h"
 
 #include <functional>
 #include <memory>
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -390,8 +390,8 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
     m_collectionNames = edmEvent.value().get().getAvailableCollections();
   } else if (m_collectionNames.empty()) {
     std::optional<std::map<uint32_t, std::string>> idToNameOpt(std::move(m_idToName));
-    auto collections = getAvailableCollectionsFromStore(this, idToNameOpt);
-    m_idToName = std::move(idToNameOpt.value());
+    auto                                           collections = getAvailableCollectionsFromStore(this, idToNameOpt);
+    m_idToName                                                 = std::move(idToNameOpt.value());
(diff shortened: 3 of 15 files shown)


k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -62,9 +61,13 @@ StatusCode Lcio2EDM4hepTool::finalize() { return AlgTool::finalize(); }
 // Check if a collection was already registered to skip it
 // **********************************
 bool Lcio2EDM4hepTool::collectionExist(const std::string& collection_name) {
-  auto collections = m_podioDataSvc->getEventFrame().getAvailableCollections();
+  std::vector<std::string> collections;
+  //TODO:
```
tmadlener, line 65: Is this still a real TODO or is it a left over?

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -208,9 +219,19 @@ StatusCode Lcio2EDM4hepTool::convertCollections(lcio::LCEventImpl* the_event) {
...
+    auto& metadataFrame = m_podioDataSvc->getMetaDataFrame();
+    for (const auto& [collName, pidInfo] : pidInfos) {
+      edm4hep::utils::PIDHandler::setAlgoInfo(metadataFrame, collName, pidInfo);
+    }
+  } else {
+    for (const auto& [collName, pidInfo] : pidInfos) {
+      k4FWCore::putParameter(podio::collMetadataParamName(collName, edm4hep::labels::PIDAlgoName), pidInfo.algoName);
+      k4FWCore::putParameter(podio::collMetadataParamName(collName, edm4hep::labels::PIDAlgoType), pidInfo.algoType());
+      k4FWCore::putParameter(podio::collMetadataParamName(collName, edm4hep::labels::PIDParameterNames),
+                             pidInfo.paramNames);
```
tmadlener, line 233: I would have a preference of doing this via https://github.com/key4hep/k4FWCore/pull/273 as we would not leak all of these details everywhere, which would also minimize the risk of breaking things accidentally because no one remembers where all the usages are.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -302,14 +310,59 @@ podio::CollectionBase* EDM4hep2LcioTool::getEDM4hepCollection(const std::string&
...
                                   lcio::LCEventImpl* lcio_event, CollectionPairMappings& collection_pairs,
                                   std::vector<EDM4hep2LCIOConv::ParticleIDConvData>& pidCollections,
                                   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>&  dQdxCollections) {
-  const auto& metadata = m_podioDataSvc->getMetaDataFrame();
-  const auto  collPtr  = getEDM4hepCollection(e4h_coll_name);
-  const auto  fulltype = collPtr->getValueTypeName();
+  std::optional<std::reference_wrapper<podio::Frame>> metadata;
+  if (m_podioDataSvc) {
+    metadata = m_podioDataSvc->getMetaDataFrame();
+  }
```
tmadlener, line 363:
```suggestion
```
I think this is unnecessary because `metadata` is only used behind the exact same check further below. Hence, we can simply get the Frame there without having to park it into an optional here.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -340,8 +393,32 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
+      maybeAlgoName = k4FWCore::getParameter<std::string>(
+          podio::collMetadataParamName(e4h_coll_name, edm4hep::labels::PIDAlgoName));
+      maybeAlgoType =
+          k4FWCore::getParameter<int>(podio::collMetadataParamName(e4h_coll_name, edm4hep::labels::PIDAlgoType));
+      maybeParamNames = k4FWCore::getParameter<std::vector<std::string>>(
+          podio::collMetadataParamName(e4h_coll_name, edm4hep::labels::PIDParameterNames));
+    }
+
+    edm4hep::utils::ParticleIDMeta pidInfo{std::move(maybeAlgoName.value()), maybeAlgoType.value(),
+                                           maybeParamNames.value()};
```
tmadlener, line 419: This should also be covered by https://github.com/key4hep/k4FWCore/pull/273, I think.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -399,13 +485,39 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
...
+      auto* frame = dynamic_cast<AnyDataWrapper<podio::Frame>*>(p);
+      edmEvent    = std::cref(frame->getData());
+    } else {
+      auto frame = podio::Frame{};
+      edmEvent   = frame;
+    }
+  }
   for (const auto& pidCollMeta : pidCollections) {
-    const auto algoId = attachParticleIDMetaData(lcio_event, edmEvent, pidCollMeta);
+    auto algoId = attachParticleIDMetaData(lcio_event, edmEvent.value(), pidCollMeta);
```
tmadlener, line 501: Maybe this also needs a similar treatment as `convertObjectParameters` in k4EDM4hep2LcioConv? That would make it possible to simplify the code below a bit, but I am not sure if it's worth the effort.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
jmcarcell: It can but then when using the `PodioDataSvc` the `pidInfo` has to be created like it is shown here so all the labels and `podio::collMetadataParamName` have to be used anyway, see a few lines above. I'm not sure what's better, at least now they are done exactly the same way when using `PodioDataSvc` and when not. Using https://github.com/key4hep/k4FWCore/pull/273 could mean the implementations could differ at some point in time...

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
jmcarcell: Done

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
jmcarcell: Yes, this is not needed

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
jmcarcell: Done

k4MarlinWrapper/CMakeLists.txt
```diff
@@ -54,6 +54,13 @@ target_include_directories(MarlinWrapper PUBLIC
   ${LCIO_INCLUDE_DIRS}
 )
 
+# k4MarlinWrapperUtils
+gaudi_add_library(k4MarlinWrapperUtils SHARED
```
andresailer, line 58: What is the reason to split this into a separate library, and not be part of the main library?

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -31,6 +33,7 @@
 #include <vector>
 
 class PodioDataSvc;
+class MetadataSvc;
```
andresailer, line 36: I don't see this `MetadataSvc` name used in the header file?

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -64,6 +67,9 @@ class EDM4hep2LcioTool : public AlgTool, virtual public IEDMConverter {
 
   PodioDataSvc*                   m_podioDataSvc;
```
andresailer, line 68: Could we have a one line description for each service under what circumstances which service is used?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -340,8 +356,14 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
   } else if (fulltype == "edm4hep::EventHeader") {
     convertEventHeader(e4h_coll_name, lcio_event);
   } else if (fulltype == "edm4hep::ParticleID") {
-    pidCollections.emplace_back(e4h_coll_name, static_cast<const edm4hep::ParticleIDCollection*>(collPtr),
-                                edm4hep::utils::PIDHandler::getAlgoInfo(metadata, e4h_coll_name));
+    std::optional<edm4hep::utils::ParticleIDMeta> pidInfo;
+    if (m_podioDataSvc) {
```
andresailer, line 360: I asked for the comments about which service is used where because of this.

k4MarlinWrapper/src/components/StoreUtils.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  SmartDataPtr<DataObject> root(thisClass->evtSvc(), "/Event");
+  if (!root) {
+    thisClass->error() << "Failed to retrieve root object /Event" << endmsg;
+  }
+
+  auto pObj = root->registry();
+  if (!pObj) {
+    thisClass->error() << "Failed to retrieve the root registry object" << endmsg;
+  }
```
tmadlener, line 49: Should we exit early here, or throw an exception to avoid running into a seg-fault? Or does `error()` already do that in this case?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -399,13 +429,39 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
...
   EDM4hep2LCIOConv::sortParticleIDs(pidCollections);
+  if (!m_podioDataSvc) {
+    DataObject* p;
+    StatusCode  code = m_eventDataSvc->retrieveObject("/Event" + k4FWCore::frameLocation, p);
+    if (code.isSuccess()) {
+      auto* frame = dynamic_cast<AnyDataWrapper<podio::Frame>*>(p);
+      edmEvent    = std::cref(frame->getData());
+    } else {
+      auto frame = podio::Frame{};
+      edmEvent   = frame;
```
tmadlener, line 441: This will lead to a dangling reference, right? The `frame` that `edmEvent` is referring to will be destroyed at the end of the `else` scope and I don't think `reference_wrapper` does any lifetime extension?

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -131,9 +146,12 @@ namespace {
 
 StatusCode Lcio2EDM4hepTool::convertCollections(lcio::LCEventImpl* the_event) {
   // Convert event parameters
-  // auto& frame = const_cast<podio::Frame&>(m_podioDataSvc->getEventFrame());
-  auto& frame = m_podioDataSvc->m_eventframe;
-  LCIO2EDM4hepConv::convertObjectParameters<lcio::LCEventImpl>(the_event, frame);
+  if (m_podioDataSvc) {
+    LCIO2EDM4hepConv::convertObjectParameters<lcio::LCEventImpl>(the_event, m_podioDataSvc->m_eventframe);
+  } else {
+    LCIO2EDM4hepConv::convertObjectParameters<lcio::LCEventImpl>(
```
tmadlener, line 152:
```suggestion
    LCIO2EDM4hepConv::convertObjectParameters(the_event, m_podioDataSvc->m_eventframe);
  } else {
    LCIO2EDM4hepConv::convertObjectParameters(
```
We don't have to specify this as it can be deduced.

jmcarcell: Fixed

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
jmcarcell: Changed to `IMetadataSvc`, which was the original intention and is used below

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
jmcarcell: This is now done after adding the new library and the function to get collections from the store when using IOSvc.

k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
jmcarcell: Like the ones I added?

k4MarlinWrapper/src/components/StoreUtils.cpp
jmcarcell: No, `error()` doesn't, however I don't know if there are any circumstances under which `error()` will run. The same code can be found in the Writer from k4FWCore and I have never seen it happening. I can add one `throw` if you really want, but then all the checks in that function should also have a `throw`.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
jmcarcell: I don't think so, `std::reference_wrapper` is basically a pointer to the object so only the pointer gets deleted: https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper. Also if that was the case, since this function is called many times and there is only one frame then it would crash when the function is called a second time.

k4MarlinWrapper/src/components/StoreUtils.cpp
tmadlener: Does a `fatal` throw or abort execution? The main potential issue I see here is that if these checks fail we will have a `nullptr` access and a segfault, and depending on if / when stdout / stderr has last been flushed we may or may not see the error message. If there are no visible error messages in case things fail, these checks are almost useless, I think. Obviously, if the messages are guaranteed to be there even in case of a seg fault, then there is at least some information where things went wrong.

jmcarcell: `fatal` doesn't abort. `endmsg` writes the message so there should always be some information and it would happen right before the crash. As I said, I've never seen this fail because I think if it's in a state where that would fail something else will fail much earlier, like not being able to find `/Event` in the store. I'll just add the 3 or 4 throws since it's easier than continuing this conversation.

jmcarcell:
I just realised I made a mistake and https://github.com/key4hep/k4EDM4hep2LcioConv/pull/102 is not needed, so it can be reverted. https://github.com/key4hep/k4EDM4hep2LcioConv/pull/102 was needed before because I was incorrectly writing the event parameters to the metadata frame. I discovered a limitation (added in a warning) of the current implementation: when running Marlin processors with `IOSvc` without any input file, it's not possible to convert event parameters to EDM4hep as the frame that will be written doesn't exist yet; it will be created later by the Writer: https://github.com/key4hep/k4FWCore/blob/main/k4FWCore/components/Writer.cpp#L200.

k4MarlinWrapper/CMakeLists.txt
andresailer: Ok, I see. Thanks!

tmadlener:
> when running Marlin processors with IOSvc without any input file, it's not possible to convert event parameters to EDM4hep as the frame that will be written doesn't exist yet; it will be created later by the Writer:

This is mainly (only?) relevant for the LCIO to EDM4hep direction, right? As far as I can tell, the tests are unchanged apart from the additional infrastructure to switch between IOSvc and PodioDataSvc. In that case I would say this is OK for now and we check again in the future if someone runs into problems with this.

jmcarcell:
I have edited the first post with information about all the tests. I think after https://github.com/key4hep/k4FWCore/pull/280 functional algorithms and IOSvc should play well with the Marlin wrapper unless I'm missing any situation that is not in the tests...

jmcarcell:
This is finally passing the CI so it's ready from my side.

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -62,12 +68,16 @@ StatusCode Lcio2EDM4hepTool::finalize() { return AlgTool::finalize(); }
...
-    if (collection_name == name) {
-      debug() << "Collection named " << name << " already registered, skipping conversion." << endmsg;
-      return true;
-    }
+  std::vector<std::string> collections;
+  if (m_podioDataSvc) {
+    collections = m_podioDataSvc->getEventFrame().getAvailableCollections();
+  } else {
+    std::optional<std::map<uint32_t, std::string>> dummy = std::nullopt;
+    getAvailableCollectionsFromStore(this, dummy, true);
```
tmadlener, line 76: For my education: Why are we not assigning the return value here to `collections`? It looks like `collections` will always be empty if we use the `IOSvc` and the `std::find` on line 78, will never find the `collection_name`?

tmadlener (commented):
I have just one minor question, but since CI is passing, I think it's just me missing something.

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
jmcarcell: I forgot to set it, now it should be fixed. This is used in the `global_converter_maps` test where `EventHeader` is skipped a few times after the first one. I see that after adding this in the output of the test it shows that is being skipped as it happens in the original test without IOSvc.

tmadlener (approved):

tmadlener:
We should also bump (or add in the first place) the minimum required version for k4FWCore in the top level CMakeLists.txt. But we need a new tag for that first for k4FWCore.

jmcarcell:
CI passed in the previous run. The nightlies don't have all the tags to make it pass until tomorrow.

PR #217, opened 2025-02-06
Use the available tag from k4EDM4hep2LcioConv

tmadlener:
BEGINRELEASENOTES
- Require the minimal version of k4EDM4hep2LcioConv to be 0.11 as that is a working tag

ENDRELEASENOTES

See: https://github.com/key4hep/k4EDM4hep2LcioConv/pull/110


tmadlener:
This doesn't seem to be fully technically necessary, but I think it's cleaner to have this properly set.

PR #221, opened 2025-02-17
Fix the name of the TrackerHitPlane map to consider them in relations properly

tmadlener:
BEGINRELEASENOTES
- Fix the `trackerHitPlanes` name of the map going from EDM4hep to LCIO to make it work properly.
- Remove some no longer necessary special handling of ParticleID and TrackerHitPlane maps.

ENDRELEASENOTES

See https://github.com/key4hep/CLDConfig/issues/64 and https://github.com/key4hep/k4EDM4hep2LcioConv/pull/112 for more details


Zehvogel (approved):
Ouch! LGTM! :+1:

Issue #219, opened 2025-02-13
convertAll ignored after changes for IOSvc compatibility?


giovannimarchiori:
### Check duplicate issues.

- [ ] Checked for duplicates

### Goal

I have a code that runs DDMarlinPandora on edm4hep files, via the edm4hep2LCIO converters:
https://gitlab.cern.ch/gmarchio/FCC-scripts/-/blob/main/run_ALLEGRO_reco.py?ref_type=heads#L972

Until today I was using for the input the legacy services  (k4DataSvc, PodioInput) because the code would not run otherwise, but I was told yesterday that with 
https://github.com/key4hep/k4MarlinWrapper/pull/216
https://github.com/key4hep/k4FWCore/pull/280
I would be able to go back to the IOSvc input.

However, when I tried this today, I get multiple warnings in output that I didn't have before, one for each hit collection that I am trying to read in DDMarlinPandora, such as:

`[ MESSAGE "DDMarlinPandora"] Failed to extract hcal calo hit collection: HCalBarrelReadoutPositioned, lcio::DataNotAvailableException: LCEventImpl::getCollection: collection not in event:HCalBarrelReadoutPositioned
`
These warnings only disappear if I modify the EDM4hepTool configuration of my code, which was:

```
    from Configurables import EDM4hep2LcioTool
    edm4hepConvTool = EDM4hep2LcioTool("EDM4hep2lcio")
    edm4hepConvTool.convertAll = True
    edm4hepConvTool.collNameMapping = {
        "MCParticles": "MCParticle",
        "TracksFromGenParticles": "TrackCollection",
    }
    pandora.EDM4hep2LcioTool = edm4hepConvTool
```
to include in the collNameMapping all the hits collections, for instance
`        "HCalBarrelReadoutPositioned": "HCalBarrelReadoutPositioned",
`

Before, this didn't seem to be necessary - it seemed to be handled automatically when convertAll = True as in my case. Is that a bug or a feature of the new changes?

Tagging @jmcarcell  and @tmadlener 

Cheers,
   Giovanni

### Operating System and Version

Rocky Linux 9

### compiler

GCC 14.2.0 (Spack GCC)

### The version of the key4hep stack

today's nightly

### Package Version

main

### Reproducer

```
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh

[...]

PR #225, opened 2025-02-28
Fix missing collections with IOSvc when using convertAll=True

jmcarcell:
BEGINRELEASENOTES
- Fix missing collections with IOSvc when using `convertAll=True`
- Add a test that would fail without this fix
- Rename `PseudoRecoAlgorithm` to `PseudoRecoFunctional` and create a new  `PseudoRecoAlgorithm` that implements a `Gaudi::Algorithm`

ENDRELEASENOTES

The issue here was that not every check that is being done in `Writer.cpp` in k4FWCore is being done since the function there only looks at the names and makes the checks later, while here it is convenient to make the checks at the same time because it is expected this way in some parts of the wrapper (the if - else to check for `PodioDataSvc` or `IOSvc` are simpler this way).

Fix https://github.com/key4hep/k4MarlinWrapper/issues/219.

### DIFF
--- a/test/gaudi_opts/test_global_converter_maps.py
+++ b/test/gaudi_opts/test_global_converter_maps.py
@@ -29,7 +29,7 @@ from Configurables import (
     Lcio2EDM4hepTool,
     EDM4hep2LcioTool,
     MCRecoLinkChecker,
-    PseudoRecoAlgorithm,
+    PseudoRecoFunctional,
     EventDataSvc,
 )
 
@@ -68,8 +68,8 @@ else:
     podioOutput = PodioOutput("OutputWriter")
     podioOutput.filename = "global_converter_maps.root"
 
-PseudoRecoAlg = PseudoRecoAlgorithm(
-    "PseudoRecoAlgorithm", InputMCs=["MCParticles"], OutputRecos=["PseudoRecoParticles"]
+PseudoRecoAlg = PseudoRecoFunctional(
+    "PseudoRecoFunctional", InputMCs=["MCParticles"], OutputRecos=["PseudoRecoParticles"]
 )
 
 inputConverter = EDM4hep2LcioTool("InputConverter")
--- a/test/gaudi_opts/test_global_converter_maps.py
+++ b/test/gaudi_opts/test_global_converter_maps.py
@@ -30,6 +30,7 @@ from Configurables import (
     EDM4hep2LcioTool,
     MCRecoLinkChecker,
     PseudoRecoFunctional,
+    PseudoRecoAlgorithm,
     EventDataSvc,
 )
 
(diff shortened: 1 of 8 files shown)


test/gaudi_opts/test_link_conversion_edm4hep.py
```diff
@@ -38,6 +39,12 @@
 parser.add_argument(
     "--iosvc", action="store_true", default=False, help="Use IOSvc instead of PodioDataSvc"
 )
+parser.add_argument(
+    "--use-gaudi-algorithm",
+    action="store_true",
+    default=False,
+    help="Use IOSvc instead of PodioDataSvc",
```
m-fila, line 46:
```suggestion
    help="Use algorithm based on Gaudi::Algorithm instead of a functional algorithm",
```

m-fila: or maybe just "Use non-functional-algorithm"?

test/gaudi_opts/test_link_conversion_edm4hep.py
```diff
@@ -74,12 +86,15 @@
...
-mcLinkConverter.convertAll = False
-mcLinkConverter.collNameMapping = {
-    "TrivialMCRecoLinks": "TrivialMCRecoLinks",
-    "MCParticles": "MCParticles",
-    "PseudoRecoParticles": "PseudoRecoParticles",
-}
+if args.use_gaudi_algorithm:
+    mcLinkConverter.convertAll = True
+else:
+    mcLinkConverter.convertAll = False
```
tmadlener, line 92: Can we remove this differentiation here? I think I originally put it in effectively as an optimization. Or would this still require https://github.com/key4hep/k4FWCore/pull/290

test/gaudi_opts/test_link_conversion_edm4hep.py
jmcarcell: Fixed

test/gaudi_opts/test_link_conversion_edm4hep.py
jmcarcell: Yes, actually when checking I thought we have `convertAll` in very places if not none so maybe it's a good idea to have some more.

jmcarcell: Actually I reverted to how it was before. I found that when `convertAll` is set, from EDM4hep to LCIO it won't convert collections that are created since it only checks which names have been passed to the converter. This could be improved by checking in the store, and we already have some functionality to do that. I'll open an issue to discuss that. We can change to always convert all later.

test/gaudi_opts/test_global_converter_maps.py
```diff
@@ -67,8 +67,8 @@
     podioOutput = PodioOutput("OutputWriter")
     podioOutput.filename = "global_converter_maps.root"
 
-PseudoRecoAlg = PseudoRecoAlgorithm(
-    "PseudoRecoAlgorithm", InputMCs=["MCParticles"], OutputRecos=["PseudoRecoParticles"]
+PseudoRecoAlg = PseudoRecoFunctional(
+    "PseudoRecoFunctional", InputMCs=["MCParticles"], OutputRecos=["PseudoRecoParticles"]
```
tmadlener, line 71: Can we also make this a "two-pronged" test? I.e. run it once with `Functional` and once with `Gaudi::Algorithm`? Otherwise we might run into similar issues as we are fixing here again at some point.

PR #227, opened 2025-03-07
Use `IOSvc` by the default in docs and examples

m-fila:
BEGINRELEASENOTES
- Update docs to use `IOSvc` instead of `k4DataSvc`. Use`IOSvc` by the default in the examples

ENDRELEASENOTES

Changing `IOSvc` from opt-in to opt-out as it's supposed to be the recommended service with the upcoming deprecation of `PodioDataSvc`. https://github.com/key4hep/k4FWCore/issues/292

The instructions in the docs are updated to work with `IOSvc`

### DIFF
--- a/doc/starterkit/k4MarlinWrapperCLIC/Readme.md
+++ b/doc/starterkit/k4MarlinWrapperCLIC/Readme.md
@@ -140,7 +140,7 @@ included with k4MarlinWrapper. Note that:
 
 - To run *clicReconstruction* with EDM4hep format, use the steering file found in the `examples` folder of k4MarlinWrapper:
 `k4MarlinWrapper/examples/clicRec_e4h_input.py` (this also gets installed to `$K4MARLINWRAPPER/examples` in Key4hep releases)
-  + Change the line where `evtsvc.input` is defined to point to the location of your input file.
+  + Change the line where `IOSvc.Input` is defined to point to the location of your input file.
   + At the bottom of the file, in the `ApplicationMgr` parameters, change `EvtMax   = 3,` to the number of events to run.
 
 This can be run in the following way.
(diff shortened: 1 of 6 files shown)


doc/starterkit/k4MarlinWrapperCLIC/Readme.md
```diff
@@ -140,7 +140,7 @@ included with k4MarlinWrapper. Note that:
 
 - To run *clicReconstruction* with EDM4hep format, use the steering file found in the `examples` folder of k4MarlinWrapper:
 `k4MarlinWrapper/examples/clicRec_e4h_input.py` (this also gets installed to `$K4MARLINWRAPPER/examples` in Key4hep releases)
-  + Change the line where `evtsvc.input` is defined to point to the location of your input file.
+  + Change the line where `IOSvc.Input` is defined to point to the location of your input file.
```
tmadlener, line 143: I am not entirely sure, but these might need changes in the [`convertMarlinSteeringToGaudi.py`](https://github.com/key4hep/k4MarlinWrapper/blob/main/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py) script to actually spit out the `IOSvc` instead of the old `k4DataSvc`.

m-fila: [convertMarlinSteeringToGaudi.py](https://github.com/key4hep/k4MarlinWrapper/blob/main/k4MarlinWrapper/scripts/convertMarlinSteeringToGaudi.py) is producing a file using the default Gaudi `EventDataSvc()` and `read = LcioEvent()` which then had to be manually converted to use `k4MarlinWrapper`. https://github.com/key4hep/key4hep-tutorials/pull/21 will update instructions to manually convert to `IOSvc`

tmadlener: :+1: Thanks for the clarification. Haven't touched this for too long it seems.

m-fila: Ah and this is actually asking to modify `k4MarlinWrapper/examples/clicRec_e4h_input.py` which is already converted to use either `IOSvc` or `k4DataSvc`

Issue #226, opened 2025-03-04
When setting convertAll, collections created by algorithms are not converted from EDM4hep to LCIO


jmcarcell:
I found that when `convertAll` is set, from EDM4hep to LCIO it won't convert collections that are created since it only checks which names have been passed to the converter. This could be improved by checking in the store, and we already have some functionality to do that.

Or was `convertAll` never meant to be used without the collection name mappings @tmadlener?

_Originally posted by @jmcarcell in https://github.com/key4hep/k4MarlinWrapper/pull/225#discussion_r1979754433_

tmadlener:
> Or was `convertAll` never meant to be used without the collection name mappings @tmadlener?

No the intention definitely was to simply convert all collections it can find in the event store. Collections that have been converted via previous conversion should not be converted again. Similarly, if an LCIO collection with the same name is already present we should also not overwrite that, and instead skip the conversion.

tmadlener:
I can reproduce this with the following assumptions:
- Using `k4DataSvc` as EventDataSvc
- Using a Functional algorithm to create collections

If I use the `IOSvc` and the normal `EventDataSvc` things work regardless of whether it's a `Gaudi::Algorithm` or a Functional that creates the collections. If I use a `Gaudi::Algorithm` and the `k4DataSvc` things also work.

I think this is quite expected, as the `Gaudi::Algorithm` effectively puts collections into the TES in such a way that the `k4DataSvc` can directly work with it and it will thus end up in the internal Frame. However, if it does not end up in the internal Frame and the `k4DataSvc` is used, the call to `getAvailableCollections` below will not contain collections created by Functional algorithms:
https://github.com/key4hep/k4MarlinWrapper/blob/8ddcfde47753cfe99319e231a9809d9ca2448fac/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp#L389-L397

PR #233, opened 2025-03-11
Ensure that all EDM4hep collections are discovered for conversion

tmadlener:
Add a new test case to showcase all possible behaviors

BEGINRELEASENOTES
- Unconditionally check the TES when looking for EDM4hep collections to convert in case `convertAll` is set to `True`.
- Restructure conversion slightly and only do this lookup once and cache the results from the first conversion. This is possible since event contents are assumed to be constant.

ENDRELEASENOTES

~This is a draft PR to showcase the behavior in https://github.com/key4hep/k4MarlinWrapper/issues/226#issuecomment-2714564618~

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -28,6 +28,7 @@
 
 #include <map>
 #include <string>
+#include <tuple>
 #include <vector>
 
 class PodioDataSvc;
@@ -70,9 +71,9 @@ private:
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
   // Metadata service from k4FWCore that is used together with IOSvc
   SmartIF<IMetadataSvc> m_metadataSvc;
-  /// A (caching) map of original to new collection names that will be populated
+  /// A (caching) "map" of original to new collection names that will be populated
   /// during the first conversion
-  std::unordered_map<std::string, std::string> m_collsToConvert{};
+  std::vector<std::tuple<std::string, std::string>> m_collsToConvert{};
   std::map<uint32_t, std::string> m_idToName;
 
   void convertTracks(TrackMap& tracks_vec, const std::string& e4h_coll_name, const std::string& lcio_coll_name,
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -70,7 +70,9 @@ private:
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
   // Metadata service from k4FWCore that is used together with IOSvc
   SmartIF<IMetadataSvc> m_metadataSvc;
-  std::vector<std::string> m_collectionNames;
+  /// A (caching) map of original to new collection names that will be populated
+  /// during the first conversion
+  std::unordered_map<std::string, std::string> m_collsToConvert{};
   std::map<uint32_t, std::string> m_idToName;
 
   void convertTracks(TrackMap& tracks_vec, const std::string& e4h_coll_name, const std::string& lcio_coll_name,
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -398,14 +398,15 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
       info() << "Converting all collections from EDM4hep to LCIO" << endmsg;
       std::vector<std::string> collectionNames{};
       if (m_podioDataSvc) {
+        // If we have the PodioDataSvc get the collections available from frame
         edmEvent = m_podioDataSvc->getEventFrame();
         collectionNames = edmEvent.value().get().getAvailableCollections();
-      } else {
-        std::optional<std::map<uint32_t, std::string>> idToNameOpt(std::move(m_idToName));
-        auto collections = getAvailableCollectionsFromStore(this, idToNameOpt);
-        m_idToName = std::move(idToNameOpt.value());
-        collectionNames.insert(collectionNames.end(), collections.begin(), collections.end());
       }
+      // Always check the contents of the TES
+      std::optional<std::map<uint32_t, std::string>> idToNameOpt(std::move(m_idToName));
+      auto collections = getAvailableCollectionsFromStore(this, idToNameOpt);
+      m_idToName = std::move(idToNameOpt.value());
+      collectionNames.insert(collectionNames.end(), collections.begin(), collections.end());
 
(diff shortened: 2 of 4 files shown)


k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
```diff
@@ -68,9 +68,11 @@ class EDM4hep2LcioTool : public AlgTool, virtual public IEDMConverter {
   // EventDataSvc that is used together with IOSvc
   ServiceHandle<IDataProviderSvc> m_eventDataSvc;
   // Metadata service from k4FWCore that is used together with IOSvc
-  SmartIF<IMetadataSvc>           m_metadataSvc;
-  std::vector<std::string>        m_collectionNames;
-  std::map<uint32_t, std::string> m_idToName;
+  SmartIF<IMetadataSvc> m_metadataSvc;
+  /// A (caching) map of original to new collection names that will be populated
+  /// during the first conversion
+  std::unordered_map<std::string, std::string> m_collsToConvert{};
```
jmcarcell, line 74:
```suggestion
  std::map<std::string, std::string> m_collsToConvert{};
```
Maybe not worth it to be an unordered_map since the number of collections is probably never going to be very big?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -386,25 +386,33 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
+
+    if (m_convertAll) {
+      info() << "Converting all collections from EDM4hep to LCIO" << endmsg;
+      std::vector<std::string> collectionNames{};
+      if (m_podioDataSvc) {
+        // If we have the PodioDataSvc get the collections available from frame
+        edmEvent        = m_podioDataSvc->getEventFrame();
+        collectionNames = edmEvent.value().get().getAvailableCollections();
+      }
+      // Always check the contents of the TES
```
jmcarcell, line 405: Always because when using the PodioDataSvc with functional algorithms the collections don't go to the event frame, otherwise this would be an `if-else` (I thought initially it should be an `if-else`).

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -386,25 +386,33 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
-    info() << "Converting all collections from EDM4hep to LCIO" << endmsg;
-    // And simply add the rest, exploiting the fact that emplace will not
-    // replace existing entries with the same key
-    for (const auto& name : m_collectionNames) {
-      collsToConvert.emplace(name, name);
+  // use m_collsToConvert to detect whether we run the first time and cache the
+  // results as we can assume that all the events have the same contents
+  if (m_collsToConvert.empty()) {
+    // Start off with the pre-defined collection name mappings
+    m_collsToConvert = {m_collNames.value().begin(), m_collNames.value().end()};
```
jmcarcell, line 393:
```suggestion
    m_collsToConvert = m_collNames.value();
```
I think changing to a map instead of an unordered map this should work?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -386,25 +386,33 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
+      if (m_podioDataSvc) {
+        // If we have the PodioDataSvc get the collections available from frame
+        edmEvent        = m_podioDataSvc->getEventFrame();
+        collectionNames = edmEvent.value().get().getAvailableCollections();
+      }
+      // Always check the contents of the TES
+      std::optional<std::map<uint32_t, std::string>> idToNameOpt(std::move(m_idToName));
+      auto                                           collections = getAvailableCollectionsFromStore(this, idToNameOpt);
+      m_idToName                                                 = std::move(idToNameOpt.value());
+      collectionNames.insert(collectionNames.end(), collections.begin(), collections.end());
```
jmcarcell, line 409: `collectionNames` is now a temporary without any purpose other than to fill `m_collsToConvert`. So it could be avoided completely by looping twice, once for the collections from the event data frame (if any) and then a second (or first) time for the collections in the store.

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
tmadlener: Yes. Should I put this info into the comment?

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
```diff
@@ -386,25 +386,33 @@ void EDM4hep2LcioTool::convertAdd(const std::string& e4h_coll_name, const std::s
...
-    // replace existing entries with the same key
-    for (const auto& name : m_collectionNames) {
-      collsToConvert.emplace(name, name);
+  // use m_collsToConvert to detect whether we run the first time and cache the
+  // results as we can assume that all the events have the same contents
+  if (m_collsToConvert.empty()) {
+    // Start off with the pre-defined collection name mappings
+    m_collsToConvert = {m_collNames.value().begin(), m_collNames.value().end()};
+    // We *always* want to convert the EventHeader
+    m_collsToConvert.emplace(edm4hep::labels::EventHeader, "<directly into LCEvent>");
```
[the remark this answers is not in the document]
tmadlener: Yes this is a bit annoying at the moment. For the moment we simply documented that it needs to be there (in #229):

image

We should probably fix this a bit better (in a different PR).

k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
tmadlener: I have made this a `vector<tuple<string, string>>` since it's only used to iterate after it has been populated initially.

test/gaudi_opts/test_global_converter_maps.py
```diff
@@ -90,11 +90,7 @@
     )
 
 inputConverter = EDM4hep2LcioTool("InputConverter")
-inputConverter.convertAll = False
-inputConverter.collNameMapping = {
-    "MCParticles": "MCParticles",
-    "PseudoRecoParticles": "PseudoRecoParticles",
-}
+inputConverter.convertAll = True
```
jmcarcell, line 93: I guess this was failing before and would show up in the `TrivialMCTruthLinker` that it would miss some collections?

tmadlener: Yes, this was failing before for the `PodioInput` version of the test, because we only read `EventHeader` and `MCParticles`, but before https://github.com/key4hep/k4FWCore/pull/290 the `Frame` in the `PodioDataSvc` would claim that it has all collections and then fail when they are not in the TES. Initially this was an optimization to make the test quicker to run by only reading and converting what is strictly necessary.

tmadlener:
Looks like CI is stable enough again to make this pass. Unless there are more comments, I will merge this tomorrow.

PR #237, opened 2025-05-21
Make sure to not double convert link collections

tmadlener:
BEGINRELEASENOTES
- Make sure that Link collections are properly skipped from conversion if they already exist in the LCIO event

ENDRELEASENOTES

- [x] Test case covering this


tmadlener:
No issue that I am aware of, just stumbled over this in #236. I need to check how much work it would be to put this into a test.

tmadlener:
I have adjusted the existing link conversion checks to include a double conversion and it indeed fails before this fix:

```
23: ToolSvc.Another...  DEBUG Adding LCRelation collection TrivialMCRecoLinks of type LCRelation
23: AnotherLinkChecker  FATAL Standard std::exception is caught in sysExecute
23: AnotherLinkChecker  ERROR lcio::EventException: LCEventImpl::addCollection() name already exists: TrivialMCRecoLinks
23: EventLoopMgr        FATAL .executeEvent(): Standard std::exception thrown by AnotherLinkChecker
23: EventLoopMgr        ERROR lcio::EventException: LCEventImpl::addCollection() name already exists: TrivialMCRecoLinks
23: EventLoopMgr      WARNING Execution of algorithm AnotherLinkChecker failed
```

PR #238, opened 2025-05-22
Refactor retrieval of EDM4hep event Frame into method

tmadlener:
BEGINRELEASENOTES
- Refactor the retrieval of the EDM4hep event (Frame) into standalone method

ENDRELEASENOTES

Another standalone outcome of #236 

As far as I can tell there is no way that the optional is not engaged that didn't already break before. The behavior before was
- If there is a PodioDataSvc, we get the event Frame from there
- Otherwise we get the Frame from where the IOSvc puts it
  - In case it is not there we create an empty local Frame

@jmcarcell anything I missed here about the previous behavior?

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/EDM4hep2Lcio.h
@@ -121,7 +121,7 @@ private:
                   std::vector<EDM4hep2LCIOConv::ParticleIDConvData>& pidCollections,
                   std::vector<EDM4hep2LCIOConv::TrackDqdxConvData>& dQdxCollections);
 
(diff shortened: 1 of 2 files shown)


tmadlener:
Playing around with godbolt, it indeed looks like the previous version only worked by accident (if it worked at all) [reproducer](https://godbolt.org/z/PjKaKoY3d).

I have switched to returning a `const &` and throwing an exception in case there is no valid Frame available. At least for our tests everything seems to work.

tmadlener:
Tests are passing for me locally. I have to investigate why they are failing here.

tmadlener:
Not sure why tests were passing for me locally, I can't seem to reproduce this. I have added the necessary functionality to put an empty Frame into the TES if none is there already.

jmcarcell (approved):
Looks good. I edited the details about removing `std::optional` and `std::reference_wrapper`, which now is done and was a bit confusing.

PR #239, opened 2025-05-23
Move duplicated code into helper function

tmadlener:
BEGINRELEASENOTES
- Refactor duplicated code to retrieve the global converted objects map into a helper function and call that instead

ENDRELEASENOTES

Another off-spring of #236

### DIFF
--- a/k4MarlinWrapper/src/components/StoreUtils.cpp
+++ b/k4MarlinWrapper/src/components/StoreUtils.cpp
@@ -99,3 +99,24 @@ std::vector<std::string> getAvailableCollectionsFromStore(const AlgTool* thisCla
   }
   return collectionNames;
 }
+
+k4MarlinWrapper::GlobalConvertedObjectsMap& getGlobalObjectMap(AlgTool* thisTool) {
+  // We want one "global" map that is created the first time it is used in the event.
+  DataObject* obj = nullptr;
+  auto sc = thisTool->evtSvc()->retrieveObject(k4MarlinWrapper::GlobalConvertedObjectsMap::TESpath.data(), obj);
+  if (sc.isFailure()) {
+    thisTool->debug() << "Creating GlobalconvertedObjectsMap for this event since it is not already in the EventStore"
+                      << endmsg;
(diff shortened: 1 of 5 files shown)


k4MarlinWrapper/src/components/StoreUtils.cpp
```diff
@@ -99,3 +99,24 @@ std::vector<std::string> getAvailableCollectionsFromStore(const AlgTool* thisCla
   }
   return collectionNames;
 }
+
+k4MarlinWrapper::GlobalConvertedObjectsMap& getGlobalObjectMap(AlgTool* thisTool) {
+  // We want one "global" map that is created the first time it is used in the event.
+  DataObject* obj = nullptr;
+  auto sc = thisTool->evtSvc()->retrieveObject(k4MarlinWrapper::GlobalConvertedObjectsMap::TESpath.data(), obj);
+  if (sc.isFailure()) {
+    thisTool->debug() << "Creating GlobalconvertedObjectsMap for this event since it is not already in the EventStore"
```
tmadlener, line 108:
```suggestion
    thisTool->debug() << "Creating GlobalConvertedObjectsMap for this event since it is not already in the EventStore"
```

k4MarlinWrapper/src/components/StoreUtils.cpp
```diff
@@ -99,3 +99,24 @@ std::vector<std::string> getAvailableCollectionsFromStore(const AlgTool* thisCla
...
+                      << endmsg;
+    auto globalObjMapWrapper = new AnyDataWrapper(k4MarlinWrapper::GlobalConvertedObjectsMap{});
+    auto nsc = thisTool->evtSvc()->registerObject(k4MarlinWrapper::GlobalConvertedObjectsMap::TESpath.data(),
+                                                  globalObjMapWrapper);
+    if (nsc.isFailure()) {
+      thisTool->error() << "Could not register GlobalConvertedObjectsMap in the EventStore" << endmsg;
```
tmadlener, line 114:
```suggestion
```

PR #246, opened 2025-09-24
Make it possible again to not convert everything

Zehvogel:
BEGINRELEASENOTES
-  Make it possible again to use the collection mapping to specify which collections should be converted

ENDRELEASENOTES

Looks like this accidentally disappeared in https://github.com/key4hep/k4MarlinWrapper/pull/233, due to counting braces on GitHub being very annoying :(

Discovered while trying to work around https://github.com/key4hep/k4EDM4hep2LcioConv/issues/122 by only converting what is needed. We should probably add a test for this :/

@tmadlener

### DIFF
--- a/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
+++ b/k4MarlinWrapper/src/components/EDM4hep2Lcio.cpp
@@ -456,10 +456,9 @@ StatusCode EDM4hep2LcioTool::convertCollections(lcio::LCEventImpl* lcio_event) {
         debug() << fmt::format("Adding '{}' from TES to conversion? {}", name, inserted) << endmsg;
       }
       m_idToName = std::move(idToNameOpt.value());
-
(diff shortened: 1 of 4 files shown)


tmadlener:
I do agree on the test for this. I think we should have enough of the bits and pieces to make a meaningful test where we have at least one `EDM4hep2LcioTool` configured with `convertAll = False`. The check could simply be a `podio-dump` or `anajob` with a `grep -v` (or maybe some `FAIL_REGULAR_EXPRESSION` in CTest) to make sure collections that weren't requested for conversion really weren't converted, or use one of the Marlin processors that expects a converted collection and let that fail with a `DataNotAvailableCollection` as success case.

Zehvogel:
@tmadlener I have added a test, but only to check that the explicitly requested collection exists. I don't really have the time for something more complicated at the moment :(

PR #253, opened 2025-10-07
Make it possible to run without EventHeader collection when converting from EDM4hep to LCIO

tmadlener:
BEGINRELEASENOTES
- Allow for missing `EventHeader` collection when converting from EDM4hep to LCIO.

ENDRELEASENOTES

This has been converted unconditionally up until now. This also means that the whole run fails if the collection is missing. Now we check (for the first event), whether we have an `EventHeader` collection and only then add it to the conversion by default even if it's unspecified by the user.

- [ ] Includes #252


tmadlener:
Merging this later today unless there are comments.

PR #254, opened 2025-10-09
Do not refresh seeds for every instance

tmadlener:
BEGINRELEASENOTES
- Refresh the seeds for Marlin exactly once instead of doing it for every wrapped processor (over and over)

ENDRELEASENOTES

@andresailer does this look like a viable solution to you? I am happy to take back the second commit otherwise.

- [x] Needs #255


tmadlener:
Removed the `thread_local` again as I don't think it is actually necessary given that we are not running concurrently in any case.

tmadlener:
I will merge this later today, so that it can be included in the next tag.

PR #472, opened 2026-01-16
Make sure we can still build with EDM4hep 1.0

tmadlener:
BEGINRELEASENOTES
- Make sure we can still build with EDM4hep 1.0

ENDRELEASENOTES


tmadlener:
CI failure should be unrelated to these changes. Merging.

PR #475, opened 2026-04-02
Remove support for legacy I/O components

tmadlener:
BEGINRELEASENOTES
- Remove support for running the wrapper with legacy I/O components since they will be removed with [k4FWCore#392](https://github.com/key4hep/k4FWCore/pull/392)

ENDRELEASENOTES

This is largely just removing things. I have moved the `getEDM4hepEvent` into the common `StoreUtils` since we need this in both directions. Otherwise it's essentially just switching to newer utilities from k4FWCore and removing / changing tests that used `k4DataSvc`.

- [x] Needs https://github.com/key4hep/k4FWCore/pull/391 (as this uses `k4FWCore::(get|put)CellIDEncoding`
- [x] Needs https://github.com/key4hep/k4FWCore/pull/390 (as we use the utilities from a `AlgTool`)

### DIFF
--- a/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
+++ b/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
@@ -50,7 +50,7 @@ Lcio2EDM4hepTool::Lcio2EDM4hepTool(const std::string& type, const std::string& n
 
 StatusCode Lcio2EDM4hepTool::finalize() {
   for (const auto& [key, value] : m_cellIDEncodings) {
-    k4FWCore::putParameter(key, value, this);
+    k4FWCore::putCellIDEncoding(key, value, this);
   }
   for (const auto& [coll, pidMeta] : m_pidMetas) {
     k4FWCore::putParameter(coll, pidMeta, this);
@@ -96,7 +96,7 @@ void Lcio2EDM4hepTool::registerCollection(
     for (auto& elem : string_keys) {
       if (elem == edm4hep::labels::CellIDEncoding) {
         const auto& lcio_coll_cellid_str = lcioColl->getParameters().getStringVal(lcio::LCIO::CellIDEncoding);
-        m_cellIDEncodings[podio::collMetadataParamName(name, edm4hep::labels::CellIDEncoding)] = lcio_coll_cellid_str;
+        m_cellIDEncodings[name] = lcio_coll_cellid_str;
       } else {
         // TODO: figure out where this actually needs to go
       }
--- a/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
+++ b/k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
@@ -52,14 +52,6 @@ Lcio2EDM4hepTool::Lcio2EDM4hepTool(const std::string& type, const std::string& n
 StatusCode Lcio2EDM4hepTool::initialize() {
   m_podioDataSvc = dynamic_cast<PodioDataSvc*>(m_eventDataSvc.get());
 
-  if (!m_podioDataSvc) {
-    m_metadataSvc = service("MetadataSvc", false);
-    if (!m_metadataSvc) {
-      error() << "Could not retrieve MetadataSvc" << endmsg;
-      return StatusCode::FAILURE;
-    }
-  }
-
   return AlgTool::initialize();
 }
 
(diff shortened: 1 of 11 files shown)


tmadlener:
doctest workflows are broken also without this, see e.g. #476 

key4hep nightlies on alma9 suffer from a missing fix in Gaudi

tmadlener:
This can probably be merged without a new tag, since the only changes since the last tag are enabling c++26 and some CI workflow changes.

tests/gaudi_opts/test_global_converter_maps.py
```diff
@@ -59,24 +50,14 @@
...
-    if not args.use_functional_checker:
-        iosvc.Output = "global_converter_maps_iosvc.root"
-    else:
-        iosvc.Output = "global_converter_maps_iosvc_functional.root"
+evtsvc = EventDataSvc("EventDataSvc")
+
+iosvc = IOSvc()
+iosvc.CollectionNames = ["EventHeader", "MCParticles"]
+if not args.use_functional_checker:
+    iosvc.Output = "global_converter_maps_iosvc.root"
```
jmcarcell:
```suggestion
    iosvc.Output = "global_converter_maps.root"
```

tests/gaudi_opts/test_global_converter_maps.py
```diff
@@ -59,24 +50,14 @@
...
 else:
-    podioInput = PodioInput("InputReader")
-    podioInput.collections = ["EventHeader", "MCParticles"]
-    podioInput.OutputLevel = INFO
-    podioOutput = PodioOutput("OutputWriter")
-    podioOutput.filename = "global_converter_maps.root"
+    iosvc.Output = "global_converter_maps_iosvc_functional.root"
```
jmcarcell:
```suggestion
    iosvc.Output = "global_converter_maps_functional.root"
```

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
```diff
@@ -92,9 +69,8 @@ void Lcio2EDM4hepTool::registerCollection(
     return;
   }
 
-  auto wrapper = new DataWrapper<podio::CollectionBase>();
-  wrapper->setData(e4hColl.release());
-
+  debug() << fmt::format("Adding collection '{}' () to the TES", name, e4hColl->getTypeName()) << endmsg;
```
jmcarcell:
```suggestion
  debug() << fmt::format("Adding collection '{}' ({}) to the TES", name, e4hColl->getTypeName()) << endmsg;
```
Otherwise the second argument won't be substituted (maybe fmt should complain somehow?).

jmcarcell (commented):
Looks good to me, I grepped and couldn't find more places where there is something to remove.

k4MarlinWrapper/src/components/Lcio2EDM4hep.cpp
tmadlener: Interesting, I thought `fmt` would have this as a hard compiler error, but it's only the case if you provide too few arguments apparently.

tmadlener:
Now we are stumbling over https://github.com/key4hep/k4FWCore/pull/400

jmcarcell:
I should really fix the downstream build, because this is visible there but there is always some repo failing...

tmadlener:
I have picked up your changes from #478 (verbatim) and added similar changes to also set the particle id meta information to `finalize`. This works for me now for the ILD standard reconstruction with EDM4hep inputs and with LCIO inputs, so at least for a non-mixed chain this seems to solve things without key4hep/k4FWCore#406

I have to try and find a test case that requires the pre-filling.

tmadlener:
@jmcarcell should we merge this to see if key4hep/k4FWCore#406 is necessary immediately or whether we might be able to push that for a bit? I am not sure if I can put together a test case that conclusively shows that pre-loading fixes potentially remaining issues.

edit: The failing CIs are due to 
- Gaudi issue(?) in Alma9 nightly
- Missing functionality in the alma9 release
- Downstream build: https://github.com/key4hep/k4Clue/pull/78

jmcarcell:
Yes, I think that just putting in `finalize()` should be the proper fix. I don't know if in https://github.com/key4hep/k4MarlinWrapper/pull/478 because it was closed when I reran the CI it didn't take into account I pushed to the branch or it was github, because there it wasn't passing after moving one of the put to finalize and commenting out the other.

tmadlener:
I think the failure in #478 might also have been that that only moves the cell id encoding parts, but the same pattern was used for the particle id meta stuff, so there was still another exception that wasn't addressed fully there.

jmcarcell:
In the branch there it is commented out

tmadlener:
https://github.com/key4hep/k4MarlinWrapper/pull/478/changes (is github acting up again? I don't see it here). In any case, we pick it up here unless we want to split out those changes into a separate PR and make this about removing the legacy bits only?

jmcarcell:
The last commit was in the branch but apparently because it was closed it never went through, weird because I remember testing this locally... Now we can see that https://github.com/key4hep/k4MarlinWrapper/pull/478 passes the Ubuntu pipeline. I've put in https://github.com/key4hep/k4MarlinWrapper/pull/478 the last commit of this branch so it's split so up to you; it's not a completely clean split since a variable that is used also for the PodioDataSvc branch also has to be removed.

tmadlener:
Split it off to have a slightly cleaner history to come back to in case it's necessary.

I have also tested this locally with CLD and ILD reconstructions (again with LCIO and EDM4hep inputs). Both run and without crashes, so I think it should also be save to merge this (after a potential review) to unblock https://github.com/key4hep/k4FWCore/pull/392 (depending on the timeline of that).

---- 
> This can probably be merged without a new tag, since the only changes since the last tag are enabling c++26 and some CI workflow changes.

This is now no longer true!

PR #476, opened 2026-04-16
Make doctest and pre-commit workflows triggerable by dispatch

tmadlener:
BEGINRELEASENOTES
- Make doctest and pre-commit workflows triggerable by dispatch

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/doctest.yaml
+++ b/.github/workflows/doctest.yaml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v3
     - name: Start container
       run: |
-        docker run -it --name CI_container --privileged -v ${GITHUB_WORKSPACE}:/Package -d ghcr.io/key4hep/key4hep-images/ubuntu24-cvmfs:latest /bin/bash
+        docker run -it --name CI_container --privileged -v ${GITHUB_WORKSPACE}:/Package -d ghcr.io/key4hep/key4hep-images/alma9-cvmfs:latest /bin/bash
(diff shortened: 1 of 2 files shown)


.github/workflows/doctest.yaml
```diff
@@ -1,6 +1,10 @@
 name: doctest-linux
 
-on: [push, pull_request]
+on:
```
tmadlener:
```suggestion
on:
```

tmadlener:
Looks like the switch to default to LCG based nightlies for Alma9 breaks things since the releases are still done with spack and we just have two entirely different software stacks. I would not expect that to work, so I have switched the READMEs to also use the same nightlies that we use for building the version of k4MarlinWrapper that we want to later use for testing.

jmcarcell:
I don't understand, is this related to the visualization issue? I may have some input since I have seen something similar. But then I don't understand, we are using one stack or the other? In particular nothing changed if you don't pass the `--lcg` when sourcing the nightlies (only the Key4hep build action uses it).

tmadlener:
Ah, I think you are right. I initially thought that the doctest also fails for basic CLIC Readme, but looking at #477 it doesn't seem to be the case. So, in the end this only needs the `workflow_dispatch` parts, but not the change to the documentation bits.

tmadlener:
The failing visualization test will be fixed with https://github.com/iLCSoft/CLICPerformance/pull/146

PR #478, opened 2026-04-22
Set the CellID encoding parameters at `finalize` in `Lcio2EDM4hep.cpp`

jmcarcell:
BEGINRELEASENOTES
- Set the CellID encoding parameters at `finalize` in `Lcio2EDM4hep.cpp`

ENDRELEASENOTES

Since due to [4c89e08c2a29b4c4b3c82bac06a59797faf2dd82](https://github.com/key4hep/k4FWCore/commit/4c89e08c2a29b4c4b3c82bac06a59797faf2dd82) in k4FWCore, it is not possible to set parameters in the event loop anymore. I left the `PodioDataSvc` path as it is, since it will be removed soon.

### DIFF
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
@@ -64,6 +64,8 @@ private:
   SmartIF<IMetadataSvc> m_metadataSvc;
   PodioDataSvc* m_podioDataSvc;
 
+  std::map<std::string, std::string> m_cellIDEncodings{};
+
   // **********************************
   // Check if a collection was already registered to skip it
   // **********************************
--- a/k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
+++ b/k4MarlinWrapper/k4MarlinWrapper/converters/Lcio2EDM4hep.h
@@ -26,6 +26,8 @@
 
 #include "k4MarlinWrapper/converters/IEDMConverter.h"
 
+#include <edm4hep/utils/ParticleIDUtils.h>
+
 #include <lcio.h>
 
 #include <map>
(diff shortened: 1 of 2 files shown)


tmadlener:
The problem is that we truly need these during the event loop. So we would probably need to front-load these. However, that would mean that we would need to discover all of them in `initialize` (e.g. by opening the LCIO file and reading the first few events to check).

The other approach is to follow what is currently done in #477 (but more general, along the discussion of https://github.com/key4hep/k4FWCore/issues/301#issuecomment-4278880945

This whole thing with the small conceptual differences start to look to me like something that we might be more easily able to solve outside the framework (like we do for getting the [*patch collections*](https://key4hep.github.io/key4hep-doc/main/how-tos/k4edm4hep2lcioconv/doc/LCIO2EDM4hep.html#example)) and simply pass in the necessary information from the outside such that we can inject it where necessary up-front, rather than figuring it out along the way (and hacking core principles of the framework while doing so).

jmcarcell:
Yes, I didn't see it because the metadata service is being used directly in a few places instead of going through `k4FWCore::put` or `get`. In CI it didn't get there because apparently the state is `running` when `finalize()` is called, and I think that is due to how the Marlin wrapper runs because I can not reproduce that. Maybe finding it in `initialize` is simpler? Only the first event would need to be checked I think. Otherwise they would have to be passed through python

tmadlener:
> Only the first event would need to be checked I think. Otherwise they would have to be passed through python

In principle yes, but then again, depending on where you start from it's possible that the first event doesn't have all the collections (because that is not enforced in LCIO), so if you want to be sure you need to check a few at least (or potentially all). So having to manually set these from python seems like an acceptable trade-off (making that population automatic can then be done in python for example, or simply reading in a file like we do for patching collections). 

Both CLD and ILD do that already in their reconstructions.

tmadlener:
I have tested the ILD reconstruction and the CLD reconstruction with the *canonical 3 ttbar events*, both with LCIO and EDM4hep inputs (and also EDM4hep & LCIO outputs for ILD). All of these options run without failure now at least. I did not check the contents in much detail, but I would assume they should be similar (if not the same) as before as there are no apparent failures.

